PR #7, opened 2022-11-21
upgrade cache-action because only v6 is whitelisted by ASF

pjfanning:


pjfanning:
Thanks @danischroeter, I made a change after you made your review. I noticed that a custom maven repo used for a restlet jar was causing the build to fail. The URL has an expired SSL cert. I found an alternative site that works. Would you be able to re-review?

nvollmar (approved):
lgtm

danischroeter (approved):
lgtm
https://restlet.talend.com/downloads/current/

Issue #6, opened 2022-11-09
Apply same scalafmt from core pekko


mdedetrich:
Currently blocked by https://github.com/scalameta/scalameta/issues/2959

PR #8, opened 2022-12-27
Update and apply scalafmt

seglo:


pjfanning (approved):

mdedetrich (changes requested):
@seglo Thanks for the PR, wasn't aware that you wanted to do a scalafmt PR. As you probably found out, there is a bug in scalafmt which meant that adding it wasn't so trivial (see https://github.com/scalameta/scalameta/issues/2959) and I actually wanted to work on this when the underlying scalafmt issue is resolved. I just recently asked on gitter when a new release is going to arrive (see https://gitter.im/scalameta/scalafmt?at=639e36d089c38775e767a00c), such a release means that the workarounds you did in https://github.com/apache/incubator-pekko-connectors/pull/8/commits/654f229022ca4242bda345ca36931bb03804c257 wont be necessary with the new scalafmt fix.

Also the PR's for scalafmt are a bit more involved, i.e. there is a very specific way of doing setting them up due to needing to add a `.git-blame-ignore-revs`, which to work needs a specific layout of 3 commits with one of them needing to have an overridden git author (see https://github.com/apache/incubator-pekko-grpc/pull/3 for an example). This means you would have to go to the trouble of rebasing your current PR into 3 commits. Also because of this you shouldn't do anything but scalafmt changes (i.e. https://github.com/apache/incubator-pekko-connectors/pull/8/commits/df123fe9032c8a6b78aade62368f245beb4337c1 should be in a separate PR). 

Do you still want to go ahead with this? I already have a flow setup to do this change, its just currently blocked in the previously mentioned PR. I apologize this wasn't clear, I should have assigned myself to https://github.com/apache/incubator-pekko-connectors/issues/6 earlier.

seglo:
Thanks @mdedetrich. I missed https://github.com/apache/incubator-pekko-connectors-kafka/issues/2 . Sorry for stepping on your toes!

I don't have any particular attachment to applying scalafmt to this repo, I only thought you or someone else hadn't got to it. I have no problem deferring to you to do this work since you have a system in place, but I don't think there's any reason to wait for https://github.com/scalameta/scalameta/issues/2959. Since it's only affecting one call site I suggest we just re-factor it and move on.

seglo:
I rebased the artifact id renaming onto `main` and copied misc non-scalafmt changes from this PR to it #10 

I updated the callsite where new scalafmt rules were failing in a separate PR #11

PR #11, opened 2023-01-05
Update GCS unmarshaller to workaround new scalafmt rules

seglo:
In preparation of applying new scalafmt rules where this PF implementation would fail to parse.


nvollmar (approved):
lgtm

PR #15, opened 2023-01-06
Add github workflow to check scala code is formatted

mdedetrich:
Similar to pekko core, replaces the sbt scalafmt check with a github actions one. Also removes the `scalafmtCheckAll`/`scalafmtSbtCheck` check because it is now unnecessary (and it also doesn't play well with IDE's).


pjfanning (approved):
lgtm

PR #19, opened 2023-01-09
non empty link-validator.yml

pjfanning:
* the current yml - with everything commented out - is failing on every run and causing a lot of noise
* eg https://github.com/apache/incubator-pekko-connectors/actions/runs/3878566368
* better to run something and re-enable the rest when we're ready


nvollmar (approved):
lgtm

PR #20, opened 2023-01-13
update contributing doc

pjfanning:
* temporarily drop the MiMa section


mdedetrich (approved):
lgtm. I was about to ask to comment rather than removing however I just noticed that markdown comments are quite verbose so this fine for me.

PR #21, opened 2023-01-16
disable fossa check

pjfanning:
* not currently working
* will need to revisit
* disable to reduce build noise in the interim


nvollmar (approved):
lgtm

PR #26, opened 2023-02-06
uptake pekko jars, update package names and config names

pjfanning:
* also removed replaced alpakka in package names because the text replace was easier this way
* org.apache.pekko package name move (instead of akka) would be better done when we uptake the pekko core libs
* temporarily enables testing of all connectors in every commit 
* test failures seem to match the pre-existing test failures - compare with https://github.com/apache/incubator-pekko-connectors/pull/41


pjfanning:
@mdedetrich I'm not looking for an immediate review of this PR - just some advice. Would it be better if I expand this PR to include the pekko core jar uptake and org.apache.pekko package renames than to wait for this PR to be reviewed on its own?

mdedetrich:
I do think its preferable to include the updated packages in both pekko core and pekko http so that we only need a single PR to change the packages from `akka.stream.alpakka.*` to `org.apache.pekko.stream.connectors.*`

mdedetrich:
Note that its likely that you will have to wait for the various renames to land in pekko-http in order to use the updated pekko core because of the usage of package private, this is the same issue that I had to deal with in pekko-http.

pjfanning:
I remember now why I didn't want to upgrade the jars and package names yet - not only does this repo depend on akka-http but also akka-grpc. By the time, pekko-grpc is ready, this PR is likely to be conflicted and hard to fix up.

I'll leave this PR for now and may end up abandoning it. It may be easier just to start again in a few weeks.

pjfanning:
@mdedetrich @nvollmar would either of you have time to review this? I know the diff is large. There are a lot of doc references to Akka/Alpakka but I want to leave them to a dedicated doc PR. The test failures match those that are broken in main branch already (see #41).

nvollmar:
Sure, waiting for this to continue switching pekko-persistence-cassandra :)

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/scaladsl/Google.scala
nvollmar, line 1: This file was removed/re-added so we'll loose the history

nvollmar (approved):
Didn't manually load each remaining diff manually, but skimmed through a big chunk and couldn't find any obvious issues. lgtm

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/scaladsl/Google.scala
pjfanning: I did the same with this file as the other files - git has some heuristic to decide which moved files are moves with some edits as opposed to files that are deleted and new ones added in their place

nvollmar: Yep, since they are fairly light I'd not consider this an issue.

PR #29, opened 2023-02-10
add disclaimer

pjfanning:
https://incubator.apache.org/guides/branding.html#disclaimers


mdedetrich (approved):
lgtm. I will update this later to use the sbt-apache-sonatype plugin I created.

PR #30, opened 2023-02-18
add apache license headers

pjfanning:


project/Common.scala
```diff
@@ -113,7 +113,5 @@ object Common extends AutoPlugin {
     Test / testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-F", "4"),
     scalafmtOnCompile := false,
     javafmtOnCompile := false,
-    headerLicense := Some(HeaderLicense.Custom(
```
mdedetrich, line 116: Would it be possible to incorporate the same changes to the header as was done in pekko, i.e. https://github.com/apache/incubator-pekko/blob/main/project/CopyrightHeader.scala ?

Just want to make sure that it works with sbt-header so that we can enforce a check to make sure that all source files are correctly formatted?

mdedetrich (changes requested):
We should do the necessary adjustments to sbt-header so that it can correctly check the headers are there.

project/Common.scala
pjfanning: this setting is in CopyrightHeader.scala - seemed better to keep the stuff together there

mdedetrich: Oh perfect then, missed it in the review.

mdedetrich (approved):
lgtm

Issue #28, opened 2023-02-06
add the Apache license headers


pjfanning:
relates to https://github.com/apache/incubator-pekko/issues/38

pjfanning:
https://github.com/apache/incubator-pekko-connectors/pull/30 covers at least some of this.

pjfanning:
PR merged

PR #31, opened 2023-02-20
add license header to conf files

pjfanning:
relates to https://github.com/apache/incubator-pekko-connectors/issues/28

### DIFF
--- a/cassandra/src/main/resources/reference.conf
+++ b/cassandra/src/main/resources/reference.conf
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0
+
 alpakka.cassandra {
   # The implementation of `akka.stream.alpakka.cassandra.CqlSessionProvider`
   # used for creating the `CqlSession`.
--- a/couchbase/src/main/resources/reference.conf
+++ b/couchbase/src/main/resources/reference.conf
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0
+
 alpakka.couchbase {
   session {
     # The couchbase nodes to connect to as hostnames of ip-addresses
--- a/google-cloud-bigquery-storage/src/main/resources/reference.conf
+++ b/google-cloud-bigquery-storage/src/main/resources/reference.conf
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0
+
 alpakka.google {
 
   credentials.default-scopes = ${?alpakka.google.credentials.default-scopes} ["https://www.googleapis.com/auth/bigquery.readonly"]
--- a/google-cloud-bigquery/src/main/resources/reference.conf
+++ b/google-cloud-bigquery/src/main/resources/reference.conf
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0
+
 alpakka.google {
   credentials.default-scopes = ${?alpakka.google.credentials.default-scopes} ["https://www.googleapis.com/auth/bigquery"]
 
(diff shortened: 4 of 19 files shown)


mdedetrich:
Question posted here https://github.com/apache/incubator-pekko-http/pull/73#issuecomment-1437681549

pjfanning:
The reference conf is probably as much part of the codebase as the scala/java source. sbt-header has built-in support for it.
Other config files are also regularly given headers in ASF projects.

some examples from Apache Spark:
* https://github.com/apache/spark/blob/master/.asf.yaml
* https://github.com/apache/spark/blob/master/scalastyle-config.xml
* https://github.com/apache/spark/blob/master/pom.xml
* https://github.com/apache/spark/blob/master/conf/fairscheduler.xml.template
* https://github.com/apache/spark/blob/master/core/src/main/resources/META-INF/services/org.apache.spark.deploy.history.EventFilterBuilder

An sbt file: https://github.com/apache/incubator-toree/blob/master/kernel/build.sbt

I've had a little trouble with license headers in .md files but this looks like the way to do it:
https://github.com/apache/druid/edit/master/docs/operations/use_sbt_to_build_fat_jar.md (note the project description block needs to start the file or Paradox will not recognise it as a description block.

I can't find any ASF use of HOCON files (probably out there, but just can't track one down) - but I can't see why a .conf should not be headered like anything else.

mdedetrich:
> I can't find any ASF use of HOCON files (probably out there, but just can't track one down) - but I can't see why a .conf should not be headered like anything else.

As a disclaimer IANAL.

The main reason is that conf files are treated completely differently as a source file since`.conf` files are expected to be edited and modified separately from source files, i.e. think of the case where someone just copies the `.conf` file to use as a base configuration and just tweaks the default values. Due to this In practice `.conf` files are treated much more like documentation then actual source files.

You then get into points about copyright, i.e. there was a reason that there wasn't any Lightbend copyright headers on any of the `*.conf` files because generally speaking none of the content inside of the config file would be considered expressive enough for copyright. You can also google this point yourself (i.e. see https://groups.drupal.org/node/17555 for a discussion on this).

pjfanning:
I am not at all convinced - I think all the conf files should have headers but we can delay this till later.

mdedetrich:
Sure, I am making a thread on the mailing list where its more appropriate to discuss then. We don't have to solve this now.

mdedetrich:
Discussion on mailing for this topic at https://lists.apache.org/thread/9plz30o8g2258jmnb9mym3bfsdfyh9pz

mdedetrich:
So we got a response from the legal ticket (see https://issues.apache.org/jira/browse/LEGAL-633?focusedCommentId=17692908&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17692908), and the recommendation was to add the header to conf files however it was suggested to use the shortened SPDX-License-Identifier (which in our case would be `Apache-2.0`) rather than the full header that is typical for source files.

pjfanning:
So something like:
```
<!--- SPDX-License-Identifier: Apache-2.0 -->
```

Based on https://spdx.org/licenses/

mdedetrich:
> So something like:
> 
> ```
> 
> ```
> 
> Based on https://spdx.org/licenses/

Yes, although specifically for `*.conf` files I think it should be

```
# SPDX-License-Identifier: Apache-2.0
```

I believe `<!---` is a HTML/XML based comment

mdedetrich:
I just made a comment on the legal ticket about whether the `SPDX-License-Identifier: ` part is necessary or not (it wasn't entirely clear in the legal ticket), see https://issues.apache.org/jira/browse/LEGAL-633?focusedCommentId=17693913&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17693913.

Once this is clarified then I will approve PR.

mdedetrich (approved):
Okay so we got a response https://issues.apache.org/jira/browse/LEGAL-633?focusedCommentId=17695876&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17695876 and indeed format is correct

lgtm

PR #33, opened 2023-03-04
fix issue with license headers

pjfanning:
* the issue solution tries to create 2 settings confs, each with different license text - 1 with full header for scala/java one 1 with SPDX text for confs
* this doesn't work as the license text set for the 2nd (the SPDX one) overwrites the first
* this solution is messy but fixes the issue by setting headerLicense only once (the full header text) and ignoring this and using the SPDX text when dealing with conf headers

### DIFF
--- a/google-cloud-bigquery-storage/src/test/resources/application.conf
+++ b/google-cloud-bigquery-storage/src/test/resources/application.conf
@@ -1,1 +1,3 @@
+# SPDX-License-Identifier: Apache-2.0
+
 akka.http.server.preview.enable-http2 = on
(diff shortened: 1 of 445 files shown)


google-cloud-bigquery-storage/src/test/resources/application.conf
```diff
@@ -1 +1,3 @@
+# SPDX-License-Identifier: Apache-2.0
+
 akka.http.server.preview.enable-http2 = on
```
mdedetrich, line 3: nit: Can we add new line here

pjfanning: 2 blanks lines before the actual configs?

pjfanning: I see now that this file doesn't end with a new line, I will add that

mdedetrich (approved):
lgtm

PR #39, opened 2023-03-10
Delete nightly-builds.yml

pjfanning:
* Remove newly added job.
* Job doesn't work without the various docker containers needed by specific modules being started (eg JMS module needs a JMS server)
* Need to look into why the existing CI build that runs for every commit is skipping these tests


mdedetrich (approved):
lgtm

PR #41, opened 2023-03-11
[DO NOT MERGE] Experiment with running full test of all connectors

pjfanning:
* Current CI only runs tests for connectors whose code has changed
* But we need a baseline of what is really working

### DIFF
--- a/.github/workflows/check-build-test.yml
+++ b/.github/workflows/check-build-test.yml
@@ -145,23 +145,15 @@ jobs:
       #  with: # gh-detect-changes.sh compares with the target branch
       #    fetch-depth: 0
 
-      - name: Detect changes in Pekko Connector ${{ matrix.connector }}
-        env:
-          CONNECTOR: ${{ matrix.connector }}
-        run: ./scripts/gh-detect-changes.sh "${CONNECTOR}"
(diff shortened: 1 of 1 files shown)


pjfanning:
@mdedetrich feel free to make changes - I have basically already merged this via another PR - the pre-existing build was just never running the tests because it decided that no PR ever had any meaningful changes.

I think for the next while, it is ok to run all the tests. It would be useful to have PR builds be more choosy. I still think we need a nightly run that runs all the test just in case.

mdedetrich:
iirc, at least when this was Alpakka all of the tests were run every PR unless I happened to miss something so I am not sure why we need a nightly to replicate this.

The tests themselves are not that heavy because they are just basic integration tests with the various services. Historically the main issue has been starting/stopping the services, which is what using testcontainers was meant to solve (related to the issue I posted earlier).

One thing we can do is to use `addSbtPlugin("com.hpe.sbt" % "sbt-pull-request-validator" % "1.0.0")` which allows you to only run tests on PR's for modules that have changed. This in conjunction with testcontainers means that PR's would only run the changed tests and we can then add a nightly job that tests everything.

PR #48, opened 2023-03-14
Fix cassandra tests

nvollmar:


nvollmar:
increasing the timeout does the trick

pjfanning (approved):
lgtm

Issue #47, opened 2023-03-14
investigate broken google-pub-sub tests


pjfanning:
see https://github.com/apache/incubator-pekko-connectors/pull/41/checks

PR #53, opened 2023-03-15
Set dynverSonatypeSnapshots to fix snapshot version

mdedetrich:
Fixes the issue with publishing locally

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,6 @@
 ThisBuild / resolvers += "Apache Nexus Snapshots".at("https://repository.apache.org/content/repositories/snapshots/")
 ThisBuild / resolvers ++= Resolver.sonatypeOssRepos("snapshots")
+ThisBuild / apacheSonatypeProjectProfile := "pekko"
 
 lazy val `pekko-connectors` = project
   .in(file("."))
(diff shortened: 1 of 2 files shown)


project/Common.scala
```diff
@@ -118,6 +117,8 @@ object Common extends AutoPlugin {
     // timeout.
     Test / testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-F", "4"),
     scalafmtOnCompile := false,
-    javafmtOnCompile := false,
-    sonatypeProfileName := "org.apache.pekko")
+    javafmtOnCompile := false)
+
+  override lazy val buildSettings = Seq(
+    dynverSonatypeSnapshots := true)
```
mdedetrich, line 123: @pjfanning This is what was causing the issue. If you dont use sbt-ci-release you need to set this value to true so that `dynverSonatypeSnapshots` sets the project version to a snapshot unless you are at a commit with a git tag.

The reason why it was publishing locally is that sbt-sonatype was actually creating a local bundle which is whats used when doing a release publish and this was happening because without `dynverSonatypeSnapshots` the projects didn't have the `-SNAPSHOT`

pjfanning (approved):
lgtm

build.sbt
```diff
@@ -1,5 +1,6 @@
 ThisBuild / resolvers += "Apache Nexus Snapshots".at("https://repository.apache.org/content/repositories/snapshots/")
 ThisBuild / resolvers ++= Resolver.sonatypeOssRepos("snapshots")
+ThisBuild / apacheSonatypeProjectProfile := "pekko"
```
mdedetrich, line 3: This change is being done just so its consistent with the rest of the projects.

pjfanning:
These tests fail consistently - https://github.com/apache/incubator-pekko-connectors/issues/47

mdedetrich:
The google pubsub failure is unrelated so merging the PR.

PR #59, opened 2023-03-15
Remove FQCN boilerplate

mdedetrich:
Remove the FQCN boilerplate so that pekko-grpc is consistent with the other pekko projects. I also checked the docs to make sure the imports render correctly.


pjfanning (approved):
lgtm

PR #65, opened 2023-03-16
Scala2.12 support

pjfanning:
S3 code needs a lot of changes to compile in Scala. 2.12

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -56,7 +56,8 @@ object Common extends AutoPlugin {
       "-deprecation",
       "-Xlint",
       "-Ywarn-dead-code",
-      "-target:jvm-1.8"),
+      "-target:jvm-1.8",
+      "-Wconf:cat=unused-nowarn:s"),
     Compile / doc / scalacOptions := scalacOptions.value ++ Seq(
       "-doc-title",
       "Apache Pekko Connectors",
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/S3Headers.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/S3Headers.scala
@@ -23,7 +23,7 @@ import pekko.stream.connectors.s3.headers.{ CannedAcl, ServerSideEncryption, Sto
 import pekko.stream.connectors.s3.impl.S3Request
 
 import scala.collection.immutable.Seq
-import scala.collection.JavaConverters._
+import org.apache.pekko.util.ccompat.JavaConverters._
 
 final class MetaHeaders private (val metaHeaders: Map[String, String]) {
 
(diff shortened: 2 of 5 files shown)


nvollmar (approved):
lgtm

mdedetrich:
At first glance if this problem is what I think it is (i.e. Scala 2.13 having a stdlib collection rewrite which changed some interfaces) then the best course of action would be to use https://github.com/scala/scala-collection-compat. The whole idea of this library is it provides common interfaces to various Scala versions to help deal with this problem.

Its also possible to just inline the abstractions if we don't want to add an additional dependency, although I would argue that scala-collection-compat is ultra stable at this point.

@pjfanning Do you want to me work on this?

pjfanning:
I'm not convinced by the need for scala-collection-compat. I've even reduced the uses of `toIndexedSeq` to the point that most have pre-existed the commit that removed Scala 2.12 support. https://github.com/apache/incubator-pekko-connectors/commit/49ec3aad562e50cf79dd5d8344956f5053400f26

mdedetrich:
I am not sure if using `immutable.Seq` is desirable here because the point of using normal `Seq` is that it also accepts mutable collections which can be relevant from the Java side of things.

Also note that other akka/pekko libraries have abstracted over these interfaces (such as `IndexedSeq`) in a similar way to scala-collection-compat but in a much more simple way and to me I think this is preferable.

I have to look at this more concretely though.

pjfanning:
I can double check but I don't think this PR affects any public APIs. The changes are in private or package private code. There is a broken unit test though. I'll debug it later.

If you want to try Scala-collection-compat, give it a go. We can see if the code that gets used is small and decide afterwards whether to inline that code or not.

mdedetrich (changes requested):
Okay so I had a proper look at the changes regarding `immutable.Seq` and it is indeed safe. This was due to another change where in Scala 2.13 they changed `Seq` to mean `immutable.Seq`.

I noticed there are some other refactoring's in the PR, would it be possible to split them out  into the PR (unless they are necessary for 2.12 which I don't think is the case?). This is firstly to keep the scope of PR's small but also in the interest of making 1.0.x as similar to Alpakka as possible incase there are some backports (completely fine with putting this change in 1.1.x).

Also the S3 tests are now failing (they werent before) and I suspect this may be due to these changes breaking a mocking test.

pjfanning:
@mdedetrich I built the PR by reverting back to the changes before Scala 2.12 was dropped and selectively adding changes. Nonetheless, I revisited the changes and added back more of the changes so the PR is smaller now.

mdedetrich (approved):
@pjfanning  Thanks for the PR. These changes are very good, we actually made the code slightly faster due to removing unnecessary conversions.

I added 2 commits onto the PR, one of them fixes on last case of a `.toIndexedSeq` which wasn't necessary (it was a workaround for forgetting to use `immutable.Seq` in another area).

The second commit I added removes all warnings due to the Scala cross compiling that we just fixed. Feel free to merge the PR if there aren't any additional changes needed.

project/Common.scala
```diff
@@ -56,7 +56,8 @@ object Common extends AutoPlugin {
       "-deprecation",
       "-Xlint",
       "-Ywarn-dead-code",
-      "-target:jvm-1.8"),
+      "-target:jvm-1.8",
+      "-Wconf:cat=unused-nowarn:s"),
```
mdedetrich, line 60: This is needed because we use the `@nowarn` annotation to suppress a warning only for Scala 2.13 and without this scalac option it would complain that no warnings were suppressed for Scala 2.12.

PR #66, opened 2023-03-16
Force US_EAST_1 for S3 listBuckets call

mdedetrich:
So this is a bug which I discovered after I added this feature (https://github.com/akka/alpakka/pull/2899) to Alpakka. I didn't manage to fix this in Alpakka because they changed their license to BSL before I discovered it.

Essentially, if you are using the list buckets call from a configured region that is not `US_EAST_1` it will fail. As noted this only occurs in AWS and not Minio (Minio is a 3rd party implementation of AWS). I have added a test to confirm this behaviour, i.e. if you revert `HttpRequests.listBuckets(s3Headers.headers), Some(Region.US_EAST_1))` to `HttpRequests.listBuckets(s3Headers.headers))` the test will fail. Unfortunately in order to replicate this you need an actual AWS account to test against (I personally used a company test account to verify this).

Regarding whether this should be put into 1.1.x or 1.0.x there are arguments both ways. In the current state the `listBuckets` call will simply just fail to work if you are using a configured non `US_EAST_1` region. The only way to work around this is to manually override the attribute (i.e. https://github.com/aiven/guardian-for-apache-kafka/blob/9fce5561479d9800d82583e68b2ff1f0191d46c6/core-s3/src/test/scala/io/aiven/guardian/kafka/s3/Main.scala#L83-L92). If you do this manual workaround in conjunction with this bug fix there is no difference, where as if you don't have this bug fix it wont work at all. So in summary while it is a behavioural change, its very obvious and hence black and white what the change is and most critically there won't be any conflicts if users implemented a manual workaround (like the one I previously mentioned). That being said I have no issues if the community agrees to target this for 1.1.x

### DIFF
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
@@ -175,12 +175,12 @@ import scala.util.{ Failure, Success, Try }
     Flow[ByteString].orElse(Source.single(ByteString.empty))
 
   // def because tokens can expire
-  private def signingKey(implicit settings: S3Settings) = {
+  private def signingKey(overrideRegion: Option[Region] = None)(implicit settings: S3Settings) = {
     val requestDate = ZonedDateTime.now(ZoneOffset.UTC)
     SigningKey(
       requestDate,
       settings.credentialsProvider,
-      CredentialScope(requestDate.toLocalDate, settings.s3RegionProvider.getRegion, "s3"))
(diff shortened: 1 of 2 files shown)


mdedetrich:
Related tests (i.e. S3) passed

pjfanning (approved):
Bug fixes make sense

pjfanning:
if it doesn't work without this fix, merge it now

mdedetrich:
Cool, I'll wait for a week or so to see if anyone else has any objections and then I will merge

mdedetrich:
So I am merging this since no one objected.

PR #68, opened 2023-03-17
Use Pekko Java/Scala converters to remove warnings

mdedetrich:
The usage of `import scala.collection.JavaConverters._` happens to create warnings when used under Scala 2.12. Thankfully pekko core happens to contain `org.apache.pekko.util.ccompat.JavaConverters._` which is designed precisely to avoid this issue (amongst others).

This should only be merged after https://github.com/apache/incubator-pekko-connectors/pull/65 is merged.


nvollmar (approved):
lgtm - we can drop them once we drop Scala 2.12

He-Pin:
I think is is fine as the first version of pekko is 1.0.0.

He-Pin:
some test if error:
```
CI / connectors (mqtt-streaming, docker-compose up -d mqtt) (pull_request) Failing after 2m

```

PR #70, opened 2023-03-20
Update pekko snapshot version and cleanup imports

mdedetrich:
Updates pekko core snapshot version which includes the latest collection compat (to avoid any potential dependency resolution issues). Also cleans up the imports which was changed in https://github.com/apache/incubator-pekko-connectors/pull/68


pjfanning (approved):
lgtm

PR #72, opened 2023-03-20
updated deprecated version to highlight that version is an Alpakka version

pjfanning:
* only 1 entry that seems to need updating


mdedetrich (approved):
lgtm

Issue #71, opened 2023-03-20
use LazyList instead of Stream in S3Stream


pjfanning:
As part of reintroducing Scala 2.12 support, we reverted the Alpakka change to use LazyList and went back to the Scala Stream class.

* we could modify the code so that Scala 2.12 code uses Stream and Scala 2.13+ uses LazyList (new in 2.13 but preferred)
* we could copy in the code from scala-collection-compat that does this or we can write our own - if we use scala-collection-compat code, we will need to update our LICENSE and NOTICE
* we could add a dependency in the s3 module to scala-collection-compat - avoids us having to update our LICENSE and NOTICE

pjfanning:
@mdedetrich this issue is about possibly using LazyList again

mdedetrich:
> @mdedetrich this issue is about possibly using LazyList again

Ah I see

One thing I want to mention is that the usage of `Stream`/`LazyList` in the first place might be an oversight since Pekko provides its own way to generate a stream of values.

In other words we might be able to just use `Source.repeat` instead of `Stream`/`LazyList.continuously`. Ontop of this `Stream`/`LazyList.continuously` is only being used in tests and the only non test code that uses `Stream`/`LazyList.continuously` is in `S3Stream` functions that are deprecated.

PR #73, opened 2023-03-21
Finish Scala 2.12 support

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko-connectors/issues/71

Finishes the Scala 2.12 support. Some more test classes needed to be changed in order to compile, pekko-connectors-kafka version was necessary to bump because a snapshot was just released last night that had 2.12 build plus some other tweaks (see inline comments)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -153,8 +153,15 @@ lazy val geode =
     "geode",
     Dependencies.Geode,
     Test / fork := true,
-    // https://github.com/scala/bug/issues/12072
-    Test / scalacOptions += "-Xlint:-byname-implicit")
+    Test / scalacOptions ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2, n)) if n == 13 =>
+          // https://github.com/scala/bug/issues/12072
+          List("-Xlint:-byname-implicit")
+        case Some((2, n)) if n == 12 =>
+          List.empty
+      }
+    })
 
 lazy val googleCommon = pekkoConnectorProject(
   "google-common",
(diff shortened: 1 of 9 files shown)


google-common/src/test/scala/docs/scaladsl/GoogleCommonDoc.scala
```diff
@@ -21,8 +21,7 @@ import pekko.stream.scaladsl.Source
 
 import scala.annotation.nowarn
 
-@nowarn("msg=never used")
-@nowarn("msg=dead code")
+@nowarn("msg=never used|dead code")
```
mdedetrich, line 24: This was necessary because for some reason the `@nowarn` annotation behaved differently on Scala 2.12 where it would complain if you had multiple `@nowarn` statements, i.e.

```
[info] compiling 1 Scala source to /Users/mdedetrich/github/incubator-pekko-connectors/google-common/target/scala-2.12/test-classes ...
[error] /Users/mdedetrich/github/incubator-pekko-connectors/google-common/src/test/scala/docs/scaladsl/GoogleCommonDoc.scala:26:7: class nowarn may not appear multiple times on class GoogleCommonDoc
[error] class GoogleCommonDoc {
```
Thankfully `msg` syntax accepts regex so this could easily be solved with regex or clause.

build.sbt
```diff
@@ -153,8 +153,15 @@ lazy val geode =
     "geode",
     Dependencies.Geode,
     Test / fork := true,
-    // https://github.com/scala/bug/issues/12072
-    Test / scalacOptions += "-Xlint:-byname-implicit")
+    Test / scalacOptions ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2, n)) if n == 13 =>
+          // https://github.com/scala/bug/issues/12072
+          List("-Xlint:-byname-implicit")
```
mdedetrich, line 160: This flag only exists for Scala 2.13, not 2.12

pjfanning (approved):
lgtm

PR #74, opened 2023-03-21
Remove usages of Stream

mdedetrich:
Completely removes usages of `Stream`, making https://github.com/apache/incubator-pekko-connectors/issues/71 redudnant.

Resolves: https://github.com/apache/incubator-pekko-connectors/issues/71

### DIFF
--- a/file/src/test/scala/docs/scaladsl/ExecutableUtils.scala
+++ b/file/src/test/scala/docs/scaladsl/ExecutableUtils.scala
@@ -18,7 +18,6 @@ import java.nio.file.{ Files, Path, Paths }
 
 import org.apache.pekko.util.ByteString
 
-import scala.annotation.nowarn
 import scala.concurrent.Future
 import scala.sys.process.{ BasicIO, Process }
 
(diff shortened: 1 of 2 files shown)


mdedetrich:
Hmm `ByteString.apply` is being a bit funny, will check this out after lunch.

file/src/test/scala/docs/scaladsl/ExecutableUtils.scala
```diff
@@ -58,10 +57,9 @@ object ExecutableUtils {
     finally stream.close()
   }
 
-  @nowarn("msg=deprecated")
   private def readStream(stream: InputStream): ByteString = {
     val reader = new BufferedInputStream(stream)
-    try ByteString(Stream.continually(reader.read).takeWhile(_ != -1).map(_.toByte).toArray)
+    try ByteString(Iterator.continually(reader.read).takeWhile(_ != -1).map(_.toByte).toArray)
```
mdedetrich, line 62: So I realized what the underlying problem was, it turns out that specifically for Scala 2.12, `ByteString` is missing the apply method for an `Iterator` where as this exists in Scala 2.13. This is likely an oversight and a PR at Pekko root was created at https://github.com/apache/incubator-pekko/pull/257 to resolve this.

When that PR lands its possible to remove the `.toArray`, in any case both variants are equivalent to eachother because internally in the `ByteString.apply` it converts it to an Array.

mdedetrich:
@pjfanning PR is now ready

nvollmar (approved):
lgtm

PR #75, opened 2023-03-21
=s3 Clean up some code and make it compiles.

He-Pin:
I think there is no need for the two subMaterialization.

### DIFF
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
@@ -1024,13 +1024,14 @@ import scala.util.{ Failure, Success, Try }
   /**
    * Initiates a multipart upload. Returns a source of the initiated upload with upload part indicess
    */
-  private def initiateUpload(s3Location: S3Location,
+  private def initiateUpload(
+      s3Location: S3Location,
       contentType: ContentType,
       s3Headers: immutable.Seq[HttpHeader]): Source[(MultipartUpload, Int), NotUsed] =
     Source
       .single(s3Location)
       .flatMapConcat(initiateMultipartUpload(_, contentType, s3Headers))
-      .flatMapConcat(r => Source.repeat(r))
+      .mapConcat(r => Iterator.continually(r))
       .zip(Source.fromIterator(() => Iterator.from(1)))
 
   private def poolSettings(implicit settings: S3Settings, system: ActorSystem) =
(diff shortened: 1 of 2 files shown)


s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/SplitAfterSize.scala
```diff
@@ -38,43 +38,46 @@ import scala.annotation.tailrec
...
+      .collect { case bs: ByteString => bs }
   }
 
   private case object NewStream
 
-  private def insertMarkers(minChunkSize: Long, maxChunkSize: Int) = new GraphStage[FlowShape[ByteString, Any]] {
-    val in = Inlet[ByteString]("SplitAfterSize.in")
-    val out = Outlet[Any]("SplitAfterSize.out")
-    override val shape = FlowShape.of(in, out)
+  private def insertMarkers(minChunkSize: Long, maxChunkSize: Int): GraphStage[FlowShape[ByteString, Any]] =
```
He-Pin, line 48: Adding an explicit type annotation, otherwise, it won't compile locally.

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
```diff
@@ -1024,13 +1024,14 @@ import scala.util.{ Failure, Success, Try }
...
    */
-  private def initiateUpload(s3Location: S3Location,
+  private def initiateUpload(
+      s3Location: S3Location,
       contentType: ContentType,
       s3Headers: immutable.Seq[HttpHeader]): Source[(MultipartUpload, Int), NotUsed] =
     Source
       .single(s3Location)
       .flatMapConcat(initiateMultipartUpload(_, contentType, s3Headers))
-      .flatMapConcat(r => Source.repeat(r))
```
mdedetrich, line 1033: What is the hypothetical problem with `Source.repeat` here?

He-Pin: The flatmapConcat will need another submaterialization

mdedetrich: Ah so its a performance improvement, 👍

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/SplitAfterSize.scala
mdedetrich: This is compiling fine for me locally and its also compiling fine in CI, do you have something unique with your setup?

mdedetrich: @He-Pin Did you manage to look into this? Otherwise the PR looks fine from my side.

PR #76, opened 2023-03-22
Update pekko and remove ByteString Scala 2.12 workaround

mdedetrich:
Updates Pekko to latest version which brings in the fix for `ByteString`, allowing us to remove the `.toArray` workaround that was introduced in https://github.com/apache/incubator-pekko-connectors/pull/74


nvollmar (approved):
lgtm

PR #80, opened 2023-04-18
Update Minio version to improve flakiness

mdedetrich:
The latest version of Minio seems to have fixed a flaky bug where the container wasn't started even though it reported itself as ready


pjfanning (approved):
lgtm

PR #81, opened 2023-04-19
Only pass SSE headers for multipart upload requests

mdedetrich:
This PR fixes https://github.com/akka/alpakka/issues/2906, Alpakka created a regression whereby their latest Alpakka BSL release (4.0.0) has multipart upload broken.

According to S3 docs (see https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html#API_CompleteMultipartUpload_Examples), you are not allowed to have any other headers that is not SSE related.

I tested this on my companies S3 account and I can confirm that it passes.

### DIFF
--- a/s3/src/test/scala/docs/scaladsl/S3SinkSpec.scala
+++ b/s3/src/test/scala/docs/scaladsl/S3SinkSpec.scala
@@ -251,6 +251,8 @@ class S3SinkSpec extends S3WireMockBase with S3ClientIntegrationSpec with Option
     val sseCSourceAlgorithmHeader = "x-amz-copy-source-server-side-encryption-customer-algorithm"
     val sseCSourceAlgorithmHeaderValue = "AES256"
     val sseCSourceKeyHeader = "x-amz-copy-source-server-side-encryption-customer-key"
+    val sseCKeyHeaderMd5 = "x-amz-server-side-encryption-customer-key-MD5"
+    val sseCKeyHeaderMd5Value = "md5"
     val sseCSourceKeyHeaderValue = sseCustomerKey
 
     val result =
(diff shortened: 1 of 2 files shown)


mdedetrich:
I have added this PR onto the 1.0.0 project because currently the Pekko S3 client is broken without this fix.

s3/src/test/scala/docs/scaladsl/S3SinkSpec.scala
```diff
@@ -286,10 +288,13 @@ class S3SinkSpec extends S3WireMockBase with S3ClientIntegrationSpec with Option
...
 
-    // No SSE-C headers required for CompleteMultipartUpload
+    // SSE headers only
     mock.verifyThat(
       postRequestedFor(urlEqualTo(s"/$targetBucketKey?uploadId=$uploadId"))
-        .withHeader(requestPayerHeader, new EqualToPattern(requestPayerHeaderValue)))
+        .withHeader(sseCAlgorithmHeader, new EqualToPattern(sseCAlgorithmHeaderValue))
+        .withHeader(sseCKeyHeader, new EqualToPattern(sseCKeyHeaderValue))
+        .withHeader(sseCKeyHeaderMd5, new EqualToPattern(sseCKeyHeaderMd5Value))
+        .withoutHeader(requestPayerHeader))
```
He-Pin, line 297: `new EqualToPattern(...)` would looks better with `EqualToPattern(...) `

mdedetrich: I just followed the same style that is done in the rest of the test

He-Pin (approved):
lgtm

PR #82, opened 2023-04-19
Only run exception thrown on existing bucket for Minio

mdedetrich:
Comment in PR has explanation


He-Pin (approved):
LGTM

mdedetrich:
@He-Pin I have a slight improvement coming up, don't merge yet

mdedetrich:
Okay I just improved the PR by also adding a test that confirms the S3 behaviour. Tested it against S3 account and I can confirm that it works.

He-Pin:
Seems the connector part is not easy to tests😁😁😁

mdedetrich:
@He-Pin 

You can read https://issues.apache.org/jira/browse/INFRA-24353?focusedCommentId=17713815&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17713815 for context but the current problem we are dealing with right now is that because these tests never ran against a real AWS account in CI a lot of tests have bit rotted (i.e. gradually broken) over time and I am now trying to fix them.

Even worse in some cases there are actual regressions and/or S3 changed their API behaviour, i.e. https://github.com/apache/incubator-pekko-connectors/pull/81

PR #83, opened 2023-04-19
Set contentCount values to 0

mdedetrich:
The fact that these values are set to 4 and 5 in the base test trait is likely complete oversight. For some context, these values are used in the `list with real credentials`, `list with real credentials using the Version 1 API`  and `list with real credentials in non us-east-1 zone` tests. In all of these cases, these are the first tests to run in the test suite and when they do run the buckets are empty which is why 0 is the correct value.

These tests were passing as expected in Minio since they were overridden to be 0 but they were failing in AWS S3.


pjfanning (approved):
lgtm - pending CI

mdedetrich:
Relevant CI passed, merging.

PR #85, opened 2023-04-29
Improve performance of S3 ByteString test generators

mdedetrich:
As mentioned in the removed TODO, this PR is improving the performance of `createStringCollectionWithMinChunkSizeRec` by keeping track of the current chunk size. This was causing problems running the tests against S3 because the current implementation was converting the entire data structure to an array and recounting it.


nvollmar (approved):
lgtm

PR #86, opened 2023-04-29
Add bucketVersioningEnabled helper methods

mdedetrich:
Follow up on from https://github.com/apache/incubator-pekko-connectors/pull/84, one of the  common usecases for bucket versioning  is to figure out if its enabled/disabled and due to the current field being nested inside of an option with a custom ADT the helper methods are being added to trivialize the operation.

Note that a Java alias is being added since the rest of the models in S3 follow the `field`/`getField` for Scala/Java repsectively.

### DIFF
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/model.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/model.scala
@@ -146,6 +146,12 @@ final class BucketVersioningResult private (val status: Option[BucketVersioningS
   def withMfaDelete(value: Boolean): BucketVersioningResult =
     copy(mfaDelete = Some(value))
 
+  /** Java API */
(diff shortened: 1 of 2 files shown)


nvollmar (approved):
lgtm

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/model.scala
```diff
@@ -194,6 +200,12 @@ final class BucketVersioning private (val status: Option[BucketVersioningStatus]
   /** Java API */
   def getMfaDelete: java.util.Optional[MFAStatus] = mfaDelete.asJava
 
+  /** Java API */
+  def getBucketVersioningEnabled: Boolean = bucketVersioningEnabled
```
pjfanning, line 204: With Java APIs, it is more common to use 'is' instead of 'get' for the names of getters that return booleans.

PR #88, opened 2023-05-01
Rename getBucketVersioningEnabled to isBucketVersioningEnabled

mdedetrich:
@pjfanning 

This is naming is more appropriate for Java API's


mdedetrich:
Actually I am going to close this PR, not because its valid but because it would be inconsistent with the rest of the codebase (i.e. https://github.com/mdedetrich/incubator-pekko-connectors/blob/8b986332794e34450e25d9a4464591816c007faa/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/model.scala#L680).

There is definitely merit in changing the naming in general but this should be done comprehensively at once and hence its likely to target 1.1.x

PR #91, opened 2023-05-02
Fix broken GooglePubSubSpec tests

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko-connectors/issues/47

Turns out the reason why the tests were failing is because at some point an overloaded method with the `host: Option[String]` parameter was added and its this specific overloaded method which is now being called in the auth tests and the mocks in the broken tests was pointing to the original non overloaded method.


mdedetrich:
Relevant tests pass

pjfanning (approved):
lgtm

PR #92, opened 2023-05-02
Fix dependency resolution with GoogleBigQueryStorage module

mdedetrich:
This is likely a result of us not using the pekko/pekko-http versions correctly due to using snapshots but nevertheless its better to pin `pekko-http` to the same version as the other pekko-http artifacts

```
[info]   java.lang.IllegalStateException: You are using version 0.0.0+4338-c98db6bd-SNAPSHOT of Akka HTTP, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. You can solve this by adding an explicit dependency on version 0.0.0+4338-c98db6bd-SNAPSHOT of the [pekko-http] artifacts to your project. Here's a complete collection of detected artifacts: (0.0.0+4298-26846a02-SNAPSHOT, [pekko-http]), (0.0.0+4338-c98db6bd-SNAPSHOT, [pekko-http-core, pekko-http-spray-json, pekko-parsing]). See also:
```

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -206,6 +206,7 @@ object Dependencies {
       "io.grpc" % "grpc-auth" % org.apache.pekko.grpc.gen.BuildInfo.grpcVersion, // ApacheV2
       "org.apache.pekko" %% "pekko-http-spray-json" % PekkoHttpVersion,
       "org.apache.pekko" %% "pekko-http-core" % PekkoHttpVersion,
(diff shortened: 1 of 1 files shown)


mdedetrich:
Relevant tests pass https://github.com/apache/incubator-pekko-connectors/actions/runs/4864707370/jobs/8674073793?pr=92

pjfanning (approved):
lgtm

PR #96, opened 2023-05-02
Integrate AWSS3IntegrationSpec with S3 Account

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko-connectors/issues/67

Firstly before reviewing its recommended to change the diff to ignore whitespace i.e. 
image
as it will make it easier to review.

This PR is about modifying the `AWSS3IntegrationSpec` so that it can run in CI against a real AWS S3 account which Apache INFRA is supporting (see https://issues.apache.org/jira/browse/INFRA-24353). Initially the idea was to do this later, but considering that the Pekko/Akka S3 client unintentionally broke (see https://github.com/apache/incubator-pekko-connectors/pull/81) the importance of this was increased, i.e. we want to make sure that the S3 client actually works whenever a release is made.

Although the current `AWSS3IntegrationSpec` can currently run against a real S3 account it was intended to be tested manually (ergo not in CI) and hence its implementation was basic and not suitable for CI usage. This means the PR implements many features which is highly desired for running in CI, i.e. the ability for each test run to generate its own S3 buckets using randomly generated bucket names that contain a prefix (in this case `pekko-connectors`). This is so that multiple instances of `AWSS3IntegrationSpec` can be run at the same time without conflicting each other. Likewise the tests will also clean up the bucket after themselves and if this happens to fail then there will be a S3 lambda will do this automatically.

Note that this behaviour of randomly generating bucket names/cleaning up is only enabled for `AWSS3IntegrationSpec`, `MinioS3IntegrationSpec` behaves the same way as it did before (i.e. it uses static bucket names). This is due to the fact that `MinioS3IntegrationSpec` uses testcontainers and hence it will spawn a unique docker container for each test run. There is an argument that since we are randomly generating bucket names some of the behaviour is not really necessary (i.e. rather than dealing with static buckets for the entire test run we can alter the test suite so that every test generates the bucket it needs itself). This would reduce the complexity of the test suite but it would make the PR harder to review and its also not necessary right now.

One final thing to note is that the test is being run as a nightly job rather than in the usual CI that triggers on PR creation. This is due to both security reasons and practical reasons, i.e. as of now the AWS credentials are being stored as github secrets and github will only expose those secrets if the creator of the PR is a maintainer of the repo or on a cronjob trigger (this is to prevent malicious actors from just creating PR's that echo/print/send the secrets). While this problem is solvable in the sense that we can only run the `AWSS3IntegrationSpec` if the credentials are available (so that at least PR's created by maintainers runs the CI), this would add quite a lot of noise to the test due to the fact that its not possible to tag an entire test suite based on a condition (annotations only work with compile time constants).

The proper way to solve this problem is to use OIDC which also allows any PR creator to run full tests on S3 (OIDC works by creating temporary credentials that only last during CI run so even if a malicious actor steals the credentials, combined with the most restrictive set of AWS permissions its very difficult for them to actually compromise anything meaningful with them).

### DIFF
--- a/s3/src/test/scala/org/apache/pekko/stream/connectors/s3/TestUtils.scala
+++ b/s3/src/test/scala/org/apache/pekko/stream/connectors/s3/TestUtils.scala
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
+ */
+
+package org.apache.pekko.stream.connectors.s3
+
+import markatta.futiles.Retry
+import org.apache.commons.lang3.StringUtils
+import org.apache.pekko
+import org.apache.pekko.stream.connectors.s3.scaladsl.S3
+import org.apache.pekko.stream.scaladsl.Sink
+import org.scalacheck.Gen
+import pekko.actor.ActorSystem
+import pekko.stream.Attributes
+
+import scala.annotation.tailrec
+import scala.concurrent.duration._
+import scala.concurrent.{ ExecutionContext, Future }
+
+object TestUtils {
+  def checkEnvVarAvailable(env: String): Boolean =
+    try
+      sys.env.get(env) match {
+        case Some(value) if StringUtils.isBlank(value) => false
+        case Some(_)                                   => true
+        case None                                      => false
+      }
+    catch {
+      case _: NoSuchElementException => false
+    }
+
+  def cleanAndDeleteBucket(bucket: String, s3Attrs: Attributes)(implicit system: ActorSystem): Future[Unit] = {
+    implicit val ec: ExecutionContext = system.dispatcher
+    for {
+      bucketVersioningResult <- S3.getBucketVersioning(bucket)(implicitly, s3Attrs)
+      deleteAllVersions = bucketVersioningResult.status.contains(BucketVersioningStatus.Enabled)
+      _ <- S3.deleteBucketContents(bucket, deleteAllVersions = deleteAllVersions).withAttributes(s3Attrs).runWith(
+        Sink.ignore)
+      multiParts <-
+        S3.listMultipartUpload(bucket, None).withAttributes(s3Attrs).runWith(Sink.seq)
+      _ <- Future.sequence(multiParts.map { part =>
+        S3.deleteUpload(bucket, part.key, part.uploadId)(implicitly, s3Attrs)
+      })
+      _ <- Retry.retryWithBackOff(
+        5,
+        100.millis,
+        throwable => throwable.getMessage.contains("The bucket you tried to delete is not empty"))(
+        S3.deleteBucket(bucket)(implicitly, s3Attrs))
+      _ = system.log.info(s"Completed deleting bucket $bucket")
+    } yield ()
+  }
+
+  /**
+   * Will return a single value from a given generator. WARNING this will block the thread
+   * until a value is retrieved so only use this for generators that have a high chance of
+   * returning a value. If a [[Gen.filter]] ends up filtering out too many values this can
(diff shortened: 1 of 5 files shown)


s3/src/test/scala/org/apache/pekko/stream/connectors/s3/TestUtils.scala
mdedetrich, line 1: These generators are directly copied from an OS company project (see https://github.com/aiven/guardian-for-apache-kafka/blob/main/core-s3/src/test/scala/io/aiven/guardian/kafka/s3/Generators.scala) and hence I can confirm that they work correctly since they have been use for over a year.

s3/src/test/scala/org/apache/pekko/stream/connectors/s3/scaladsl/Generators.scala
mdedetrich, line 1: These generators are directly copied from an OS company project (see https://github.com/aiven/guardian-for-apache-kafka/blob/main/core-s3/src/test/scala/io/aiven/guardian/kafka/s3/Generators.scala) and hence I can confirm that they work correctly since they have been use for over a year.

s3/src/test/scala/org/apache/pekko/stream/connectors/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -48,30 +50,180 @@ trait S3IntegrationSpec
...
+    val (bucketWithDots, defaultBucket, bucketWithVersioning, nonExistingBucket) = {
+      val baseGen = for {
+        bucketWithDots <- genBucketName(S3IntegrationSpec.BucketWithDots)
+        defaultBucket <- genBucketName(S3IntegrationSpec.DefaultBucket)
+        bucketWithVersioning <- genBucketName(S3IntegrationSpec.BucketWithVersioning)
+        nonExistingBucket <- genBucketName(S3IntegrationSpec.NonExistingBucket)
+      } yield (bucketWithDots, defaultBucket, bucketWithVersioning, nonExistingBucket)
+
+      if (randomlyGenerateBucketNames)
+        TestUtils.loopUntilGenRetrievesValue(baseGen.filter {
```
mdedetrich, line 152: The `TestUtils.loopUntilGenRetrievesValue` is necessary here because we are manually invoking the scalacheck generator rather than using the Scalatest property checking integration (due to the fact its occurring in an `afterAll` block rather than an actual test).

Due to this we and the fact that we are doing a `.filter` on the generator, we cannot use the standard `sample.get` because it will return `None` if the filter fails, hence why `TestUtils.loopUntilGenRetrievesValue` exists.

s3/src/test/scala/org/apache/pekko/stream/connectors/s3/scaladsl/S3IntegrationSpec.scala
```diff
@@ -309,144 +460,161 @@ trait S3IntegrationSpec
...
+          .withAttributes(attributes)
+          .runWith(Sink.ignore)
+        result <- S3.listObjectVersions(defaultBucket, None).withAttributes(attributes).runWith(Sink.seq)
+        _ <- S3.deleteObject(defaultBucket, objectKey).withAttributes(attributes).runWith(Sink.ignore)
+      } yield result.flatMap { case (versions, _) => versions }
 
-    forEvery(results.futureValue) { version =>
-      version.versionId shouldEqual None
-    }
+      Inspectors.forEvery(results.futureValue) { version =>
```
mdedetrich, line 612: This is because the `forEvery` from Scalatest conflicts with the `forEvery` from scalacheck (which is being introduced in this PR).

mdedetrich:
@pjfanning @nvollmar Would it be possible to look into this? I want to merge outstanding PR's before I start looking into https://github.com/apache/incubator-pekko/pull/281 since merging https://github.com/apache/incubator-pekko/pull/281 will essentially block all downstream modules.

Also Apache INFRA is waiting for a response on this.

pjfanning (approved):
lgtm

PR #101, opened 2023-05-10
Apply compat changes from latest Pekko

mdedetrich:
The only notable change here is that I had to move a single source file `docs.javadsl.CassandraSessionSpec` to `org.apache.pekko.stream.connectors.cassandra.javadsl.CassandraSessionSpec`. This is because the new compat `FutureConverters`/`OptionConverters` functions are package private within `org.apache.pekko`.

Normally source files that reside within `docs.javadsl`/`docs.scaladsl` are meant to be referenced in paradox docs, but for `CassandraSessionSpec` this isn't the case. It just appears to be a standard test so it being put into `docs.javadsl` in the first place was likely an oversight/accident anyways.


mdedetrich:
@pjfanning @nvollmar @He-Pin PR is now ready, all of the dependencies have also been updated

nvollmar (approved):
lgtm

PR #103, opened 2023-05-10
Don't run problematic tests in AWS S3 account

mdedetrich:
So if you follow the `#asfinra` slack discussion at https://the-asf.slack.com/archives/CBX4TSBQ8/p1683672460539449 we managed to get most of the AWS S3 tests passing in Apache's S3 account however a few of them turned out to be problematic (see https://github.com/apache/incubator-pekko-connectors/actions/runs/4931691903/jobs/8813985233 for more info). The problematic tests were ones that involved the `ListAllMyBuckets` call + MFA, which is difficult to get right with restricted permissions (necessary for security since this Apache S3 account is being used by others).

This PR makes tests that require these permissions to be configurable as well as disabling them in our nightly CI run. While it may be theoretically possible to come up with an IAM policy to test everything, I don't have capacity for this now and the current state is good enough


pjfanning (approved):
lgtm

PR #106, opened 2023-05-16
re-enable link-validator workflow

pjfanning:
* job runs weekly but fails - I think it's because of the old ubuntu version
* the workflow has been neutered - to not run the actual validations
* so this PR re-enables them


nvollmar (approved):
lgtm

PR #108, opened 2023-05-19
Publish nightly docs

mdedetrich:
Taken from pekko-management which is published https://github.com/apache/incubator-pekko-management/blob/main/.github/workflows/publish-nightly.yml#L40-L61

### DIFF
--- a/.github/workflows/publish-nightly.yml
+++ b/.github/workflows/publish-nightly.yml
@@ -31,3 +31,26 @@ jobs:
           NEXUS_USER: ${{ secrets.NEXUS_USER }}
           NEXUS_PW: ${{ secrets.NEXUS_PW }}
         run: sbt +publish
+
+      - name: Build Documentation
+        run: |-
+          sbt docs/paradox unidoc
+
+      # Create directory structure upfront since rsync does not create intermediate directories otherwise
+      - name: Create nightly directory structure
+        run: |-
+          mkdir -p target/nightly-docs/docs/pekko-connectors/${{ github.ref_name }}-snapshot/
+          mv docs/target/paradox/site/main/ target/nightly-docs/docs/pekko-connectors/${{ github.ref_name }}-snapshot/docs
+          mv target/scala-2.13/unidoc target/nightly-docs/docs/pekko-connectors/${{ github.ref_name }}-snapshot/api
+
+      - name: Upload nightly docs
+        uses: ./.github/actions/sync-nightlies
+        with:
+          upload: true
+          switches: --archive --compress --update --delete --progress --relative
+          local_path: target/nightly-docs/./docs/pekko-connectors/ # The intermediate dot is to show `--relative` which paths to operate on
+          remote_path: ${{ secrets.NIGHTLIES_RSYNC_PATH }}/pekko/
+          remote_host: ${{ secrets.NIGHTLIES_RSYNC_HOST }}
+          remote_port: ${{ secrets.NIGHTLIES_RSYNC_PORT }}
+          remote_user: ${{ secrets.NIGHTLIES_RSYNC_USER }}
+          remote_key: ${{ secrets.NIGHTLIES_RSYNC_KEY }}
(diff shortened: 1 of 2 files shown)


pjfanning:
Is it possible that this project builds with scala-2.13 by default unlike pekko-management - https://github.com/apache/incubator-pekko-connectors/blob/bd0e1a9506bb29c08f8395bbf11b23ef9c7e709a/project/Common.scala#L50

mdedetrich:
> Is it possible that this project builds with scala-2.13 by default unlike pekko-management -
> 
> https://github.com/apache/incubator-pekko-connectors/blob/bd0e1a9506bb29c08f8395bbf11b23ef9c7e709a/project/Common.scala#L50

Aside from consistency, I don't think the default version matters that much anymore at least ever since https://github.com/apache/incubator-pekko/pull/281 and https://github.com/apache/incubator-pekko/pull/257 have been merged and propagated through the various Pekko modules.

We can make a decision on which default version of Scala should be used within our projects (I would opt for 2.13)

pjfanning:
My point is that the scalaVersion used in the build affects which dir that the build files end up. The sbt command used in this build does not specify which scala version to use so the default is used. The default appears to be 2.13 so the target dir will be `target/scala-2.13`. The workflow changes in this PR seem to use `target/scala-2.12`.

mdedetrich:
> My point is that the scalaVersion used in the build affects which dir that the build files end up. The sbt command used in this build does not specify which scala version to use so the default is used. The default appears to be 2.13 so the target dir will be `target/scala-2.13`. The workflow changes in this PR seem to use `target/scala-2.12`.

Ah I see, in that case I have updated the default Scala version to 2.12 since that is what we seem to use everywhere else. I also checked the github workflows to see if this has any unintended consequences and all seems to be good.

project/Common.scala
```diff
@@ -47,7 +47,7 @@ object Common extends AutoPlugin {
     projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
     crossVersion := CrossVersion.binary,
     crossScalaVersions := Dependencies.ScalaVersions,
-    scalaVersion := Dependencies.Scala213,
+    scalaVersion := Dependencies.Scala212,
```
pjfanning, line 50: I would prefer to use Scala213 as the default. All I'm asking for is that we github workflow uses `target/scala-2.13`

pjfanning: main pekko repos default to scala 2.13 - it's only a few like pekko-management that default to 2.12.

mdedetrich: Okay sure, ill revert this and update pekko management

mdedetrich: Done and pushed.

pjfanning (approved):
lgtm

.github/workflows/publish-nightly.yml
```diff
@@ -31,3 +31,26 @@ jobs:
...
+      - name: Build Documentation
+        run: |-
+          sbt docs/paradox unidoc
+
+      # Create directory structure upfront since rsync does not create intermediate directories otherwise
+      - name: Create nightly directory structure
+        run: |-
+          mkdir -p target/nightly-docs/docs/pekko-connectors/${{ github.ref_name }}-snapshot/
+          mv docs/target/paradox/site/main/ target/nightly-docs/docs/pekko-connectors/${{ github.ref_name }}-snapshot/docs
+          mv target/scala-2.13/unidoc target/nightly-docs/docs/pekko-connectors/${{ github.ref_name }}-snapshot/api
```
He-Pin, line 44: Maybe we can define a variable in YAML for `defaultScalaVersion`?

mdedetrich: The proper way to solve this issue is via an sbt plugin akin to https://github.com/lightbend/sbt-publish-rsync, there is already an issue on this https://github.com/apache/incubator-pekko/issues/104

He-Pin (approved):
lgtm

PR #122, opened 2023-05-29
upgrade logback

pjfanning:
* testkit project is internal
* the logback-classic dependency is duplicated
* logback 1.2.11 is what we use in pekko core when testing
* there are CVEs in logback 1.2.3 -- see https://mvnrepository.com/artifact/ch.qos.logback/logback-classic/1.2.3 (transitive dependency on logback-core 1.2.3 is the issue)


mdedetrich (approved):
lgtm

PR #123, opened 2023-05-29
initial work on scala 3 support

pjfanning:
* add Scala3 support for some of the connectors
* some like the Slick connector are stuck because there is no Scala 3 jar for that dependency
* some more might be fixed if we upgrade to a jar dependency version that supports Scala3 - but I'm wary about upgrading too many dependency versions
* there are quite a few that just don't compile and with some work they could be fixed
* https://github.com/apache/incubator-pekko-connectors/issues/126

### DIFF
--- a/amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/impl/AmqpReplyToSinkStage.scala
+++ b/amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/impl/AmqpReplyToSinkStage.scala
@@ -28,7 +28,7 @@ import scala.concurrent.{ Future, Promise }
  * the queue named in the replyTo options of the message instead of from settings declared at construction.
  */
 @InternalApi
-private[amqp] final class AmqpReplyToSinkStage(settings: AmqpReplyToSinkSettings)
+private[amqp] final class AmqpReplyToSinkStage(replyToSinkSettings: AmqpReplyToSinkSettings)
     extends GraphStageWithMaterializedValue[SinkShape[WriteMessage], Future[Done]] { stage =>
 
   val in = Inlet[WriteMessage]("AmqpReplyToSink.in")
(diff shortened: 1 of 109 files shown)


pjfanning:
@mdedetrich I'm going to pause development on this.

I think you might be looking at some of the problems already.

I think there is some value in not necessarily getting every connector working with scala 3 just yet. Start with the easiest ones and do the others over time.

mdedetrich:
@pjfanning It would be ideal to get every connector ready but yes it shouldn't block us. Thankfully we still have time since connector is down the track of our module dependency tree.

I think similar to pekko-http, there is probably merit in creating a specific Scala 3 branch as it will be easier for multiple people to work with, wdyt?

pjfanning:
I'm not sure a Scala 3 branch is that useful. I think we should just merge the different PRs to main when they are ready.

mdedetrich:
> I'm not sure a Scala 3 branch is that useful. I think we should just merge the different PRs to main when they are ready.

I would disagree here because unless Scala 3 works perfectly fine for all integrations we are going to dirty the branch which can cause issues later down the track if have to do modifications to Pekko connectors. iirc This is actually the same reason why the scala 3 support was in a separate branch for both akka-http and pekko-http, its because integrating Scala 3 support for the projects in question was complicated and long winding and due to the large amount of integrations I think this project passes that bar.

For example the immediate problem I have is that while I think merging this is fine (from what I can tell its just doing the obligatory scala source changes just so that Scala 3 compiler can accept it) there are other more complicated changes, i.e. migrating from shapless to its new Scala 3 equivalent. If we have enough of such changes then the argument for creating a Scala 3 branch increases over time because we just end up having a lot of quite non trivial changes just to support Scala 3.

In summary having a scala 3 branch allows for more iterative style development which I think is actually preferable for such a project.

pjfanning:
@mdedetrich I created a scala3 branch in the apache project. If you think this is ok as a starting point, could you approve merging this to that branch?

pjfanning:
merged to scala3 branch

PR #124, opened 2023-05-29
fix issue with Scala 2.12 testing

pjfanning:
I looked at https://github.com/apache/incubator-pekko-connectors/actions/runs/5114976247/jobs/9196215819?pr=121 and you can note that the Scala 2.12 tests are compiling to target/2.13.

I've changed the build files and using `++2.12.17` works as it should.


mdedetrich (approved):
lgtm

PR #125, opened 2023-05-30
fix up deprecation versions so they refer to Alpakka

pjfanning:
Pekko Connectors versions will start at 1.0.0 so it will be confusing if we have deprecations that are since '3.0.0'. We need to prefix such version numbers to make it clear they relate to Alpakka releases.


mdedetrich:
I am wondering whether we should use `Akka Alpakka` instead of just `Alpakka` since the `Alpakka` name itself doesn't immediately make it clear its coming from Akka (unlike the other Akka modules which always had Akka in the name?)

mdedetrich:
Fair enough, ill approve

mdedetrich (approved):
lgtm

PR #132, opened 2023-05-30
avro4s 4.1.1 (test dependency)

pjfanning:
a test only dependency - v5 is scala 3 only but I want to get up to v4 first - and then add logic to use different versions depending on scala version


nvollmar (approved):
lgtm

PR #143, opened 2023-06-08
[DRAFT] Scala3 support for more google connectors

pjfanning:


mdedetrich:
Is `.toString`an alias for `compactString`. Also the `Anyref with JsonFormat` seems a bit suspicious.

pjfanning:
> Is `.toString`an alias for `compactString`. Also the `Anyref with JsonFormat[T]` seems a bit suspicious, does it work with just `JsonFormat[T]`?

toString is an alias for compactString. The toString caused issues in some tests where the generated classes in the tests overrode the toString.

mdedetrich (approved):
lgtm

Issue #126, opened 2023-05-30
Scala 3 support


pjfanning:
* we have a real mixed bag of connectors in this project
* some are easy to make work with Scala3
* others have dependencies that don't support Scala3 or where the dependency version we use is old and there may be compatibility issues with upgrading
* some cases involve getting the code to compile in Scala3 - some looks pretty tricky - other code is easy to get working

We may be able to get some connectors Scala 3 ready for pekko connectors 1.0.0.

Currently PRs are being merged to the scala3 branch (not main).

See https://github.com/apache/incubator-pekko-connectors/labels/scala3 for other issues - where work is delayed till later.

nvollmar:
Getting the ones we easily can might already be enough for the other downstream projects to continue.

pjfanning:
@nvollmar #123 gets about half of them working but doesn't cover some of the most popular ones like the cassandra one.

PR #145, opened 2023-06-08
Scala3 support for jms connector

pjfanning:
Part of https://github.com/apache/incubator-pekko-connectors/issues/126

Scala 3 compile issues needed fixing.
* main ones were related to using protected methods on self types
* also some issues where fields were ambiguous (multiple inheritance in Scala means that a field can be declared on 2 separate traits)

I introduced a GraphStageCompanion trait to give public access to the methods that were needed to get Scala3 compile to work.

### DIFF
--- a/jms/src/main/scala/org/apache/pekko/stream/connectors/jms/impl/GraphStageCompanion.scala
+++ b/jms/src/main/scala/org/apache/pekko/stream/connectors/jms/impl/GraphStageCompanion.scala
@@ -9,8 +9,10 @@
 
 package org.apache.pekko.stream.connectors.jms.impl
 
-import org.apache.pekko.stream.Materializer
-import org.apache.pekko.stream.connectors.jms.Destination
+import org.apache.pekko
+import pekko.annotation.InternalApi
+import pekko.stream.Materializer
+import pekko.stream.connectors.jms.Destination
 
 import scala.concurrent.duration.FiniteDuration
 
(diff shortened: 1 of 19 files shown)


jms/src/main/scala/org/apache/pekko/stream/connectors/jms/impl/GraphStageCompanion.scala
```diff
@@ -0,0 +1,34 @@
...
+ * Exposes some protected methods from [[pekko.stream.stage.GraphStage]]
+ * that are not accessible when using Scala3 compiler.
+ */
+@InternalApi
+private[impl] trait GraphStageCompanion {
+  def graphStageMaterializer: Materializer
+
+  def graphStageDestination: Destination
+
+  def scheduleOnceOnGraphStage(timerKey: Any, delay: FiniteDuration): Unit
```
mdedetrich, line 29: Can we avoid using `Any` here?

mdedetrich (changes requested):
Some more cases of `Any`

jms/src/main/scala/org/apache/pekko/stream/connectors/jms/impl/GraphStageCompanion.scala
pjfanning: This signature is taken from the function I am delegating to.

pjfanning: @mdedetrich is this answer ok? The timerKey can be any case class, the 'Any' type is actually what makes sense. This is also an internal class, not an API one.

mdedetrich: Let me check this now, we really shouldn't have `Any` like this so something fishy is going on.

mdedetrich (approved):
Okay I had a deeper look and you are right, it also uses `Any` with the current version (i.e. `scheduleOnce`). This probably should be looked into but its not relevant to Scala 3

mdedetrich (approved):
lgtm

PR #146, opened 2023-06-08
scala 3 support for json streaming

pjfanning:
Part of https://github.com/apache/incubator-pekko-connectors/issues/126

A strange Scala 3 issue means that immutable.Queue.enqueue(ByteString) won't compile. I tried to produce a minimal case in scastie to report a Scala compiler bug but the minimal case works.

For now, I think using a mutable.Queue is ok and I can log an issue to keep checking when we upgrade Scala 3 to see if the issue is fixed.

### DIFF
--- a/json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala
+++ b/json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala
@@ -50,7 +50,7 @@ private[pekko] final class JsonStreamReader(path: JsonPath) extends GraphStage[F
           new JsonPathListener {
             override def onValue(value: Any, context: ParsingContext): Unit = {
               // see https://github.com/lampepfl/dotty/issues/17946
-              buffer = buffer.enqueue(Iterable.single(ByteString(value.toString)))
+              buffer = QueueHelper.enqueue(buffer, ByteString(value.toString))
             }
           })
         .build
--- a/json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala
+++ b/json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala
@@ -48,8 +48,10 @@ private[pekko] final class JsonStreamReader(path: JsonPath) extends GraphStage[F
       private val config = surfer.configBuilder
         .bind(path,
           new JsonPathListener {
-            override def onValue(value: Any, context: ParsingContext): Unit =
-              buffer = buffer.enqueue(ByteString(value.toString))
+            override def onValue(value: Any, context: ParsingContext): Unit = {
+              // see https://github.com/lampepfl/dotty/issues/17946
+              buffer = buffer.enqueue(Iterable.single(ByteString(value.toString)))
+            }
           })
         .build
       private val parser = surfer.createNonBlockingParser(config)
--- a/build.sbt
+++ b/build.sbt
@@ -256,7 +256,16 @@ lazy val ironmq = pekkoConnectorProject(
 
 lazy val jms = pekkoConnectorProject("jms", "jms", Dependencies.Jms)
 
-lazy val jsonStreaming = pekkoConnectorProject("json-streaming", "json.streaming", Dependencies.JsonStreaming)
+val scalaReleaseSeparateSource: Def.SettingsDefinition = Compile / unmanagedSourceDirectories ++= {
+  if (scalaVersion.value.startsWith("2")) {
+    Seq((LocalRootProject / baseDirectory).value / "src" / "main" / "scala-2")
+  } else {
+    Seq((LocalRootProject / baseDirectory).value / "src" / "main" / "scala-3")
+  }
+}
+
+lazy val jsonStreaming = pekkoConnectorProject("json-streaming", "json.streaming",
+  Dependencies.JsonStreaming ++ scalaReleaseSeparateSource)
 
 lazy val kinesis = pekkoConnectorProject("kinesis", "aws.kinesis", Dependencies.Kinesis)
 
--- a/json-streaming/src/main/scala-2/org/apache/pekko/stream/connectors/json/impl/QueueHelper.scala
+++ b/json-streaming/src/main/scala-2/org/apache/pekko/stream/connectors/json/impl/QueueHelper.scala
@@ -1,0 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
+ */
+
+package org.apache.pekko.stream.connectors.json.impl
+
+import org.apache.pekko.util.ByteString
+
+import scala.collection.immutable.Queue
+
+private[impl] object QueueHelper {
+  @inline final def enqueue(queue: Queue[ByteString], byteString: ByteString): Queue[ByteString] =
+    queue.enqueue(byteString)
+}
--- a/json-streaming/src/main/scala-3/org/apache/pekko/stream/connectors/json/impl/QueueHelper.scala
+++ b/json-streaming/src/main/scala-3/org/apache/pekko/stream/connectors/json/impl/QueueHelper.scala
@@ -1,0 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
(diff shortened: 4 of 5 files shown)


nvollmar (approved):
lgtm

He-Pin:
@pjfanning What's the error looks like.

mdedetrich:
Yeah I am a bit wary of changing the underlying collection type from `immutable` to `mutable`. @pjfanning Can you post the error that you got?

pjfanning:
```
[error] -- [E007] Type Mismatch Error: /Users/pj.fanning/code/incubator-pekko-connectors/json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala:52:48 
[error] 52 |              buffer = buffer.enqueue(ByteString(value.toString))
[error]    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |                    Found:    org.apache.pekko.util.ByteString
[error]    |                    Required: Iterable[org.apache.pekko.util.ByteString]
[error]    |
[error]    | longer explanation available when compiling with `-explain`
[warn] three warnings found
[error] one error found
[error] (json-streaming / Compile / compileIncremental) Compilation failed
```

There is a deprecated enqueue(Iterable[T]) function and for some reason Scala 3.3.0 compiler ignores the fact enqueue(T) is available.

mdedetrich:
> There is a deprecated enqueue(Iterable[T]) function and for some reason Scala 3.3.0 compiler ignores the fact enqueue(T) is available.

What happens if you put `ByteString(value.toString)` into a `List` with a single value, i.e.

`buffer.enqueue(List(ByteString(value.toString)))`

?

mdedetrich (changes requested):
So I just checked the code and using a mutable queue to get around this problem is fine since its internal however there is one thing I noticed, can we change `private var buffer = Queue.empty[ByteString]` to a `val` (since there is no reason for it to be a var, we are not overriding the reference since we are using a mutable collection).

pjfanning:
> So I just checked the code and using a mutable queue to get around this problem is fine since its internal however there is one thing I noticed, can we change `private var buffer = Queue.empty[ByteString]` to a `val` (since there is no reason for it to be a var, we are not overriding the reference since we are using a mutable collection).

There is a 2nd place where the queue is recreated. I can see if I can clear the original queue instead of resetting the var.

mdedetrich:
> There is a 2nd place where the queue is recreated. I can see if I can clear the original queue instead of resetting the var.

You should be able to just clear the queue?

pjfanning:
> > There is a 2nd place where the queue is recreated. I can see if I can clear the original queue instead of resetting the var.
> 
> You should be able to just clear the queue?

* clearing a queue affects the iterator that was previously created for the queue (before the clear) - causes an IndexOutOfBoundsException
* I'm trying a different solution at the moment

pjfanning:
@mdedetrich could you review the latest change when you get a chance?

json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala
```diff
@@ -69,8 +69,8 @@ private[pekko] final class JsonStreamReader(path: JsonPath) extends GraphStage[F
         }
 
         if (buffer.nonEmpty) {
-          emitMultiple(out, buffer)
-          buffer = Queue.empty[ByteString]
+          emitMultiple(out, buffer.toSeq.iterator)
```
mdedetrich, line 72: So I just noticed this now and I am thinking that the `List(ByteString(value.toString))` variant is better because while there may be a trivial box for a single `List` (which might get inlined anyways), the `buffer.toSeq.iterator` is basically going to copy the entire collection which is much worse especially considering that the size of this collection can be quite large

pjfanning:
> @pjfanning Ah I noticed the core problem, just use `enqueueAll` instead of `enqueue` (otherwise the solution is exactly the same)! `ByteString` is actually an iterator/collection and so `enqueueAll` is what we should be using, the definition of both of these functions should make it clear
> 
> ```scala
>   @deprecated("Use `enqueueAll` instead of `enqueue` to enqueue a collection of elements", "2.13.0")
>   @`inline` final def enqueue[B >: A](iter: scala.collection.Iterable[B]) = enqueueAll(iter)
> 
>   /** Creates a new queue with all elements provided by an `Iterable` object
>     *  added at the end of the old queue.
>     *
>     *  The elements are appended in the order they are given out by the
>     *  iterator.
>     *
>     *  @param  iter        an iterable object
>     */
>   def enqueueAll[B >: A](iter: scala.collection.Iterable[B]): Queue[B] = appendedAll(iter)
> ```
> 
> In other words the current version is calling `def enqueue[B >: A](iter: scala.collection.Iterable[B]) = enqueueAll(iter)` which is deprecated in 2.13 and `ByteString` is a subtype of `scala.collection.Iterable`.

That doesn't compile in Scala 2. 

```
[error] /Users/pj.fanning/code/incubator-pekko-connectors/json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala:52:52: type mismatch;
[error]  found   : org.apache.pekko.util.ByteString
[error]  required: Iterable[org.apache.pekko.util.ByteString]
[error]               buffer = buffer.enqueueAll(ByteString(value.toString))
[error]                                                    ^
[error] one error found
```

buffer is Queue[ByteString] so only ByteStrings can be queued. ByteString might be an Iterable[Byte] but that is not compatible with Queue[ByteString].

json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala
```diff
@@ -42,14 +42,16 @@ private[pekko] final class JsonStreamReader(path: JsonPath) extends GraphStage[F
...
 
       private val surfer = JsonSurferJackson.INSTANCE
       private val config = surfer.configBuilder
         .bind(path,
           new JsonPathListener {
-            override def onValue(value: Any, context: ParsingContext): Unit =
-              buffer.enqueue(ByteString(value.toString))
+            override def onValue(value: Any, context: ParsingContext): Unit = {
+              // a strange Scala 3 compiler issue means buffer.enqueue(ByteString(value.toString)) does not compile
+              buffer = buffer.enqueueAll(Seq(ByteString(value.toString)))
```
mdedetrich, line 53: I just had a look `Iterable.single` exists, its faster than `Seq` and can be passed into `buffer.enqueueAll` so `Iterable.single(Iterable.single(ByteString(value.toString)))` .

Can you try this? If it works then the PR should be fine.

json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/MyByteString.scala
```diff
@@ -0,0 +1,5 @@
+package org.apache.pekko.stream.connectors.json.impl
```
mdedetrich, line 1: This is an accident?

json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/JsonStreamReader.scala
mdedetrich: @pjfanning You can revert this line now since buffer is immutable

mdedetrich (approved):
Perfect, approving assuming CI passes.

pjfanning:
> @pjfanning Just noticed that `Iterable.single` doesn't exist in Scala 2.12. I will fix this in Pekko.

I'm trying to add a QueueHelper with different Scala 2 and Scala 3 implementations in the json-streaming module.

json-streaming/src/main/scala/org/apache/pekko/stream/connectors/json/impl/MyByteString.scala
pjfanning: yes - has been removed

pjfanning:
> > @pjfanning Just noticed that `Iterable.single` doesn't exist in Scala 2.12. I will fix this in Pekko.
> 
> I'm trying to add a QueueHelper with different Scala 2 and Scala 3 implementations in the json-streaming module.

@mdedetrich I added the QueueHelper

mdedetrich:
> I'm trying to add a QueueHelper with different Scala 2 and Scala 3 implementations in the json-streaming module.

I have already added the `Iterable.single` method to Scala 2.12 under pekko's ccompat module. The PR is here https://github.com/apache/incubator-pekko/pull/378

mdedetrich:
Okay so this solution is nicer in the sense that for Scala 2 its going to be faster/unchanged and only Scala 3 has the workaround. Still approved from me end.

pjfanning:
> Okay so this solution is nicer in the sense that for Scala 2 its going to be faster/unchanged and only Scala 3 has the workaround. Still approved from me end.

My solution has the benefit that in Scala 2, no wrapping Iterable is used

PR #154, opened 2023-06-09
support scala3 for amqp connector

pjfanning:
main issue is that Scala3 is stricter about accessing class fields during recursion

part of #126


He-Pin (approved):
lgtm

PR #159, opened 2023-06-10
support scala3 for google-cloud-storage connector

pjfanning:
small compile problem in one class (scala 3 only)

needed to add back some code accidentally reverted in https://github.com/apache/incubator-pekko-connectors/commit/8674ff591172b62986617d6eb61a4ff43a92d417

### DIFF
--- a/google-cloud-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/Formats.scala
+++ b/google-cloud-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/Formats.scala
@@ -18,7 +18,7 @@ import java.time.OffsetDateTime
 import org.apache.pekko
 import pekko.http.scaladsl.model.{ ContentType, ContentTypes }
 import pekko.stream.connectors.googlecloud.storage._
-import spray.json.{ DefaultJsonProtocol, JsObject, JsValue, RootJsonFormat, RootJsonReader }
+import spray.json.{ enrichAny, DefaultJsonProtocol, JsObject, JsValue, RootJsonFormat, RootJsonReader }
 
 import scala.util.Try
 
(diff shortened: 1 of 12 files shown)


google-cloud-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/Formats.scala
```diff
@@ -110,8 +110,8 @@ object Formats extends DefaultJsonProtocol {
       StorageObjectJson(readOnlyFields, writeableFields)
     }
     override def write(obj: StorageObjectJson): JsValue = {
-      val fields1 = obj.readable.toJson.asJsObject.fields
-      val fields2 = obj.writeable.toJson.asJsObject.fields
+      val fields1 = spray.json.enrichAny(obj.readable).toJson.asJsObject.fields
+      val fields2 = spray.json.enrichAny(obj.writeable).toJson.asJsObject.fields
```
He-Pin, line 114: What's the error here?

pjfanning: won't compile - it says that `.toJson` is not a method on the relevent classes. The implicit conversion `enrichAny` is not happening - so I have to do it explicitly.

Scala 3 compiler, unfortunately, still has some bugs.

He-Pin: how about import the implicit conversation in scope explicitly，will that work？

pjfanning: I tried basically everything. The code I have here is what the compiler generates in Scala 2 anyway.

pjfanning: I'll try again

He-Pin: wired, let's me try it too💖💖

pjfanning: looks like you were right, sorting out the imports fixed the problem - strange how Scala 2 compiler finds the implicit much more easily than Scala compiler - but it's good to know that the Scala3 compiler doesn't have a bug

He-Pin (approved):
lgtm

Issue #152, opened 2023-06-09
support Scala3 for Geode connector


pjfanning:
This will require a fair amount of rewriting because Geode connector uses the shapeless library and not the bits that are supported in shapeless3. Scala 3 Tuples will be useful.

Splitting this out (from #126) because the work is not small.

pjfanning:
I think @mdedetrich has done research on this. This issue is not regarded as a priority right now.

PR #157, opened 2023-06-09
upgrade grpc plugin

pjfanning:
Seems tidiest to upgrade to newer snapshot.

It is contributing to some issues migrating the connectors that need grpc to support scala 3.

PR #160, opened 2023-06-10
Add Scala3 support for Geode

mdedetrich:
tbd

Resolves: https://github.com/apache/incubator-pekko-connectors/issues/152

### DIFF
--- a/docs/src/main/paradox/geode.md
+++ b/docs/src/main/paradox/geode.md
@@ -69,7 +69,9 @@ Java
 :   @@snip [snip](/geode/src/test/java/docs/javadsl/PersonPdxSerializer.java) { #person-pdx-serializer }
 
 
-This Apache Pekko Connectors Geode provides a generic solution for Scala users based on [Shapeless](https://github.com/milessabin/shapeless) which may generate serializers for case classes at compile time.
+This Apache Pekko Connectors Geode provides a generic solution for Scala users based on [Shapeless](https://github.com/milessabin/shapeless) for Scala 2 and
+for Scala 3 using the [built-in tuple generic metaprogramming](https://www.scala-lang.org/2021/02/26/tuples-bring-generic-programming-to-scala-3.html)
+which may generate serializers for case classes at compile time.
 
 Java users need to implement custom serializers manually, or use runtime reflection as described in @extref[Using Automatic Reflection-Based PDX Serialization](geode:/developing/data_serialization/auto_serialization.html).
 
--- a/geode/src/main/scala-3/org/apache/pekko/stream/connectors/geode/impl/pdx/ObjectEncoder.scala
+++ b/geode/src/main/scala-3/org/apache/pekko/stream/connectors/geode/impl/pdx/ObjectEncoder.scala
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.connectors.geode.impl.pdx
+
+import org.apache.pekko.annotation.InternalApi
+
+@InternalApi
+private[pekko] trait ObjectEncoder {
+
+  given emptyTupleEncoder: PdxEncoder[EmptyTuple] =
+    PdxEncoder.instance[EmptyTuple] { case _ => true }
+
+  given tupleEncoder[K <: String, H, T <: Tuple](using
+      m: ValueOf[K],
+      hEncoder: PdxEncoder[H],
+      tEncoder: PdxEncoder[T]): PdxEncoder[FieldType[K, H] *: T] =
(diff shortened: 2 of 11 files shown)


pjfanning:
I've rebased scala3 branch so this branch will need to be rebased (eventually). https://github.com/apache/incubator-pekko-connectors/pull/157 will lead to another round of rebases.

mdedetrich:
> I've rebased scala3 branch so this branch will need to be rebased (eventually). #157 will lead to another round of rebases.

I rebased just now, just ping me when the other round of rebases is done.

pjfanning:
this is not the only connector where we have not got scala3 support - see https://github.com/apache/incubator-pekko-connectors/issues?q=is%3Aissue+is%3Aopen+label%3Ascala3

I would favour supporting Scala3 in most connectors for the 1.0.0 release of pekko-connectors but I have no problem in not supporting it in a few cases - ie I would not like to not support Scala3 just because a few connectors are posing problems.

mdedetrich:
Yup, it would be nice if we could release pekko-connectors with Scala 3 support for all connectors but its not worth it to aggressively push for it. I have a suspicion that adding support to the other connectors should be less problematic than this one as it does require more esoteric shapeless/Scala 3 knowledge to get it over the line (as opposed to other connectors which are more mechanical and/or dependency related work).

geode/src/main/scala-3/org/apache/pekko/stream/connectors/geode/impl/pdx/ObjectEncoder.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.connectors.geode.impl.pdx
+
+import org.apache.pekko.annotation.InternalApi
+
+@InternalApi
+private[pekko] trait ObjectEncoder {}
```
mdedetrich, line 23: This is still tbd

mdedetrich:
@pjfanning FYI I have updated the header for the Scala 3 specific implementations because these are not based off the Akka ones (its being done from scratch and also its using an entirely mechanism, i.e. not based off shapeless).

If you don't agree with this I can revert it.

geode/src/main/scala-3/org/apache/pekko/stream/connectors/geode/impl/pdx/ObjectEncoder.scala
jrudolph: Do you think, this is harder to do than the decoder part? Or did you not yet get to it?

mdedetrich: I did not get to it. I am not expecting any reason for it to be harder.

The main problem here personally is that my understanding of Shapeless is pretty surface level and Scala 3 has completely redone how metaprogramming works (as you can see they basically put shapeless into the core language with different abstractions).

mdedetrich: If you want to take a shot at it please go ahead, feel free to add commits onto my PR

geode/src/main/scala-3/org/apache/pekko/stream/connectors/geode/impl/pdx/Field.scala
mdedetrich, line 1: I managed to find this technique from https://scalac.io/blog/inline-your-boilerplate-harnessing-scala3-metaprogramming-without-macros/

mdedetrich:
So I managed to make some decent progress, I figured out the biggest issue I was having before is not having a type level data structure to represent the field's/types of a case class. With shapeless there is a `FieldType` however there didn't seem to be an equivalent for Scala 3's inbuild generics.

Thankfully I came across https://scalac.io/blog/inline-your-boilerplate-harnessing-scala3-metaprogramming-without-macros/ which provided the exact data structure needed, i.e.

```scala
import scala.compiletime.*
import scala.deriving.*
sealed trait Field[Label <: String, Type]

object Field {
  type FromLabelsAndTypes[Labels <: Tuple, Types <: Tuple] <: Tuple =
    (Labels, Types) match {
      case (EmptyTuple, EmptyTuple) => EmptyTuple
      case (labelHead *: labelTail, typeHead *: typeTail) =>
        Field[labelHead, typeHead] *: FromLabelsAndTypes[labelTail, typeTail]
    }
}
```

You can then create this type level data structure using a mirror, i.e.

```scala
case class Test(string: String, int: Int)

val mirror = summon[Mirror.ProductOf[Test]]

type Fields = FromLabelsAndTypes[mirror.MirroredElemLabels, mirror.MirroredElemTypes]
// defined alias type Fields = (Field["string", String], Field["int", Int])
```

Which is actually whats needed, i.e a type with each element being a `Field` that contains the case class fieldname on the left and the case class type on the right.

The PR isn't completely done but this a big step. Pinging @jrudolph because you showed interest before.

mdedetrich:
PR is ready, many thanks to @mrdziuban who helped out with the translation from shapeless to Scala 3's new tuple generic programming.

docs/src/main/paradox/geode.md
```diff
@@ -69,7 +69,9 @@ Java
 :   @@snip [snip](/geode/src/test/java/docs/javadsl/PersonPdxSerializer.java) { #person-pdx-serializer }
 
 
-This Apache Pekko Connectors Geode provides a generic solution for Scala users based on [Shapeless](https://github.com/milessabin/shapeless) which may generate serializers for case classes at compile time.
+This Apache Pekko Connectors Geode provides a generic solution for Scala users based on [Shapeless](https://github.com/milessabin/shapeless) for Scala 2 and
+for Scala 3 using the [inbuilt tuple generic metaprogramming](https://www.scala-lang.org/2021/02/26/tuples-bring-generic-programming-to-scala-3.html)
```
pjfanning, line 73: built-in is better than inbuilt

mdedetrich: Fixed and pushed.

pjfanning (approved):
lgtm

PR #161, opened 2023-06-10
Scala3 mqtt streaming

pjfanning:
part of #126 

The main problems
* Scala 3 is stricter about private fields/functions - some tests are in `docs.scaladsl` and somehow Scala 2 lets that have access to private fields - Scala 3 cares about this - I had to make some stuff public and mark them as `@InternalApi`
* MqttFlowSpec had to be changed. Scala 3 does not allow access to some stuff on self types - including implicits
  * since it is test code and it was easy to get rid of the self type, that seemed the easiest solution (I had tried a number of other ways without success)

### DIFF
--- a/mqtt-streaming/src/main/scala/org/apache/pekko/stream/connectors/mqtt/streaming/model.scala
+++ b/mqtt-streaming/src/main/scala/org/apache/pekko/stream/connectors/mqtt/streaming/model.scala
@@ -53,7 +53,8 @@ object ControlPacketType {
   val DISCONNECT = ControlPacketType(14)
   val Reserved2 = ControlPacketType(15)
 }
-final case class ControlPacketType private (underlying: Int) extends AnyVal
+@InternalApi
+final case class ControlPacketType(underlying: Int) extends AnyVal
 
 /**
  * 2.2.2 Flags
@@ -75,7 +76,8 @@ object ControlPacketFlags {
   val RETAIN = ControlPacketFlags(1)
 }
 
-final case class ControlPacketFlags private (underlying: Int) extends AnyVal {
+@InternalApi
+final case class ControlPacketFlags(underlying: Int) extends AnyVal {
 
   /**
    * Convenience bitwise OR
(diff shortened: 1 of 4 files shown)


mqtt-streaming/src/main/scala/org/apache/pekko/stream/connectors/mqtt/streaming/model.scala
```diff
@@ -53,7 +53,8 @@ object ControlPacketType {
   val DISCONNECT = ControlPacketType(14)
   val Reserved2 = ControlPacketType(15)
 }
-final case class ControlPacketType private (underlying: Int) extends AnyVal
+@InternalApi
+final case class ControlPacketType(underlying: Int) extends AnyVal
```
He-Pin, line 57: private[connectors] ?

mqtt-streaming/src/test/scala/docs/scaladsl/MqttFlowSpec.scala
```diff
@@ -36,27 +36,26 @@ import org.scalatest.matchers.should.Matchers
...
       "typed-flow-spec/topic1",
-      org.apache.pekko.actor.typed.ActorSystem(Behaviors.ignore, "TypedMqttFlowSpec").toClassic)
-    with MqttFlowSpec
-
-class ParametrizedTestKit(val clientId: String, val topic: String, system: ActorSystem) extends TestKit(system)
+      pekko.actor.typed.ActorSystem(Behaviors.ignore, "TypedMqttFlowSpec").toClassic)
 
-trait MqttFlowSpec extends AnyWordSpecLike with Matchers with BeforeAndAfterAll with ScalaFutures with LogCapturing {
-  self: ParametrizedTestKit =>
+abstract class MqttFlowSpecBase(val clientId: String, val topic: String, system: ActorSystem) extends TestKit(system)
```
He-Pin, line 48: `val` can be removed.

mqtt-streaming/src/main/scala/org/apache/pekko/stream/connectors/mqtt/streaming/model.scala
pjfanning: the code is used in a test that is in the `docs.scaladsl` package - so the private[connectors] is invisible to that package

mqtt-streaming/src/test/scala/docs/scaladsl/MqttFlowSpec.scala
pjfanning: done

He-Pin (approved):
lgtm

PR #165, opened 2023-06-10
enable scala3 build for more google connectors

pjfanning:
part of #126

### DIFF
--- a/google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
@@ -17,7 +17,6 @@ import java.time.Duration
 import java.util.concurrent.{ CompletableFuture, CompletionStage }
 import org.apache.pekko
 import pekko.actor.Cancellable
-import pekko.japi.function
 import pekko.stream.{ Attributes, Materializer }
 import pekko.stream.javadsl.{ Flow, Keep, Sink, Source }
 import pekko.{ Done, NotUsed }
@@ -62,7 +61,7 @@ object GooglePubSub {
           .setStreamAckDeadlineSeconds(0)
           .build()
 
-        val streamingPullResult: Source[StreamingPullResponse, NotUsed] = subscriber(mat, attr).client
+        subscriber(mat, attr).client
           .streamingPull(
             Source
               .single(request)
@@ -70,12 +69,10 @@ object GooglePubSub {
                 Source
                   .tick(Duration.ZERO, pollInterval, subsequentRequest)
                   .mapMaterializedValue(cancellable.complete(_))))
-
-        streamingPullResult
-          .mapConcat(new function.Function[StreamingPullResponse, java.util.List[ReceivedMessage]] {
-            override def apply(response: StreamingPullResponse): java.util.List[ReceivedMessage] =
-              response.getReceivedMessagesList
-          })
+          .mapConcat(
+            ((response: StreamingPullResponse) =>
+                  response.getReceivedMessagesList): pekko.japi.function.Function[StreamingPullResponse,
+              java.util.List[ReceivedMessage]])
           .mapMaterializedValue(_ => cancellable)
       }
       .mapMaterializedValue(flattenCs(_))
@@ -103,10 +100,10 @@ object GooglePubSub {
           .mapAsync(1, client.pull(_))
 
         val concatResult: Source[ReceivedMessage, CompletableFuture[Cancellable]] = sourceResult
-          .mapConcat(new function.Function[PullResponse, java.util.List[ReceivedMessage]] {
-            override def apply(response: PullResponse): java.util.List[ReceivedMessage] =
-              response.getReceivedMessagesList
-          })
+          .mapConcat(
+            ((response: PullResponse) =>
+                  response.getReceivedMessagesList): pekko.japi.function.Function[PullResponse,
+              java.util.List[ReceivedMessage]])
           .mapMaterializedValue(cancellable.complete(_))
           .mapMaterializedValue(_ => cancellable)
 
--- a/google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
+++ b/google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
@@ -23,7 +23,6 @@ import pekko.stream.javadsl.{ Flow, Keep, Sink, Source }
 import pekko.{ Done, NotUsed }
 import com.google.pubsub.v1._
 
-
 /**
  * Google Pub/Sub Pekko Stream operator factory.
  */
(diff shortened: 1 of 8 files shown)


mdedetrich (approved):
lgtm

google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
```diff
@@ -61,15 +62,20 @@ object GooglePubSub {
...
             Source
               .single(request)
               .concat(
                 Source
                   .tick(Duration.ZERO, pollInterval, subsequentRequest)
                   .mapMaterializedValue(cancellable.complete(_))))
-          .mapConcat(_.getReceivedMessagesList)
+
+        streamingPullResult
+          .mapConcat(new function.Function[StreamingPullResponse, java.util.List[ReceivedMessage]] {
```
He-Pin, line 75: Can this be a SAM?

pjfanning: Apologies, I'm not familiar with the term 'SAM'. Could you give me the full name?

pjfanning: In Scala3, the autoconversion of the Scala function to a Pekko JAPI Function does not happen. To be frank, I'm not really sure how the Scala2 code does this conversion but whatever it is doing does not happen automatically in Scala2 - so the code fails to compile.

mdedetrich: > Apologies, I'm not familiar with the term 'SAM'. Could you give me the full name?

SAM = Single Abstract Method, terminology introduced in Java 1.8 which means a class with just a single method aka an anonymous function. It allows the boilerplate reduction of having to manually create a class with a method (which is whats happening here).

What @He-Pin is saying is that you can probably replace

```scala
new function.Function[StreamingPullResponse, java.util.List[ReceivedMessage]] {
override def apply(response: StreamingPullResponse): java.util.List[ReceivedMessage] =
              response.getReceivedMessagesList
          })
```

With just

```scala
(response: StreamingPullResponse) => response.getReceivedMessagesList
```

pjfanning: thanks @mdedetrich - I did try that SAM format but Scala3 compiler continued to fail to compile the code

I'm afraid that Scala3 compiler doesn't always support stuff that works in Scala2 compiler.

mdedetrich: Yeah this is likely due to a regression introduced in Scala 3 where it does type inference slightly differently than Scala 2 especially when dealing with code defined in Java (i.e. `java.util.List`).

I am checking out the branch now, usually you can fix this with some manual type ascriptions.

pjfanning:
```
[error] -- [E007] Type Mismatch Error: /Users/pj.fanning/code/incubator-pekko-connectors/google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala:75:21 
[error] 75 |          .mapConcat((response: StreamingPullResponse) => response.getReceivedMessagesList)
[error]    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |         Found:    com.google.pubsub.v1.StreamingPullResponse =>
[error]    |           java.util.List[com.google.pubsub.v1.ReceivedMessage]
[error]    |         Required: org.apache.pekko.japi.function.Function[
[error]    |           com.google.pubsub.v1.StreamingPullResponse, ? <: Iterable[T]]
[error]    |
[error]    |         where:    T is a type variable with constraint 
[error]    |
[error]    | longer explanation available when compiling with `-explain`
```

The issue is that the SAM code is the same syntax as a Scala lambda function. Whatever is autoconverting the Scala function to a Java function when using Scala2 compiler just does not happen when Scala3 compiler is used.

mdedetrich: Indeed I am aware, we had similar issues when trying to put Scala 3 into Kafka. Figuring this out now

mdedetrich: @pjfanning This version works for all versions of Scala (including Scala 3)

```scala
.mapConcat(
  ((response: StreamingPullResponse) =>
        response.getReceivedMessagesList): pekko.japi.function.Function[StreamingPullResponse,
    java.lang.Iterable[ReceivedMessage]])
```

Issue appears to be due to the fact that these are custom function interfaces that Pekko has defined, I will report this upstream.

mdedetrich: Reported upstream at https://contributors.scala-lang.org/t/better-type-inference-for-scala-send-us-your-problematic-cases/2410/183. You can also add this link as a comment in the code.

google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
```diff
@@ -95,7 +98,10 @@ object GooglePubSub {
         Source
           .tick(Duration.ZERO, pollInterval, request)
           .mapAsync(1, client.pull(_))
-          .mapConcat(_.getReceivedMessagesList)
+          .mapConcat(
```
mdedetrich, line 101: Can we add a TODO/comment here that points to https://contributors.scala-lang.org/t/better-type-inference-for-scala-send-us-your-problematic-cases/2410/183 

Its likely that this will be fixed in a future version of Scala 3 at which point it would get backported to Scala 3.3.x series and we can remove this workaround.

google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
pjfanning: thanks - uptaken

google-cloud-pub-sub-grpc/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/javadsl/GooglePubSub.scala
pjfanning: done

Issue #166, opened 2023-06-10
support Scala3 for S3 connector


pjfanning:
Main blocker is that https://github.com/johanandren/futiles has no Scala 3 support. This lib is used in the test code.

fyi @mdedetrich - you appear to be a contributor to 'futiles'

mdedetrich:
> Main blocker is that https://github.com/johanandren/futiles has no Scala 3 support. This lib is used in the test code.
> 
> fyi @mdedetrich - you appear to be a contributor to 'futiles'

Yes that is true however it is maintained by an Akka/Lightbend employer which is why I avoided touching it so far.


I might be overthinking this, nothing physically is stopping me from adding Scala 3 support to it and if it gets ignored then I can always fork it.

pjfanning:
@mdedetrich I've copied the code that we need in https://github.com/apache/incubator-pekko-connectors/pull/167

PR #167, opened 2023-06-10
add scala3 support for s3

pjfanning:
relates to #166 

* have grabbed some code from futiles lib to remove the dependency (which does not yet support scala 3)
   * this will need to acknowledged in our license - futiles is Apache licensed
* Scala 3 is stricter about private access and it looks like we'll need to make some pekko-http classes public.
  * `Content-Length`   
  * `Content-Type`
* char `.toHexString` fails to compile due to https://github.com/lampepfl/dotty/issues/14630 
* A couple more compile problems also

### DIFF
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
@@ -1390,9 +1390,9 @@ import scala.util.{ Failure, Success, Try }
         import mat.executionContext
         Sink
           .seq[UploadPartResponse]
-          .mapMaterializedValue { responseFuture: Future[immutable.Seq[UploadPartResponse]] =>
+          .mapMaterializedValue { (responseFuture: Future[immutable.Seq[UploadPartResponse]]) =>
             responseFuture
-              .flatMap { responses: immutable.Seq[UploadPartResponse] =>
+              .flatMap { (responses: immutable.Seq[UploadPartResponse]) =>
                 val successes = responses.collect { case r: SuccessfulUploadPart => r }
                 val failures = responses.collect { case r: FailedUploadPart => r }
                 if (responses.isEmpty) {
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
@@ -29,7 +29,7 @@ import pekko.http.scaladsl.{ ClientTransport, Http }
 import pekko.stream.connectors.s3.BucketAccess.{ AccessDenied, AccessGranted, NotExists }
 import pekko.stream.connectors.s3._
 import pekko.stream.connectors.s3.impl.auth.{ CredentialScope, Signer, SigningKey }
-import pekko.stream.scaladsl.{ Flow, Keep, RetryFlow, RunnableGraph, Sink, Source, Tcp }
+import pekko.stream.scaladsl.{ Flow, Keep, RetryFlow, RunnableGraph, Sink, Source, SubFlow, Tcp }
 import pekko.stream.{ Attributes, Materializer }
 import pekko.util.ByteString
 import pekko.{ Done, NotUsed }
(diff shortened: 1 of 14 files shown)


mdedetrich:
Since this is only used for testing I am fine with this approach

mdedetrich:
I'll review this when I get back home and have laptop

pjfanning:
> I'll review this when I get back home and have laptop

@mdedetrich would you have time to review this?

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
```diff
@@ -1177,11 +1177,15 @@ import scala.util.{ Failure, Success, Try }
 
         import conf.multipartUploadSettings.retrySettings._
 
-        SplitAfterSize(chunkSize, chunkBufferSize)(atLeastOneByteString)
-          .via(getChunkBuffer(chunkSize, chunkBufferSize, maxRetries)) // creates the chunks
-          .mergeSubstreamsWithParallelism(parallelism)
+        val source1: SubFlow[Chunk, NotUsed, Flow[ByteString, ByteString, NotUsed]#Repr, Sink[ByteString, NotUsed]] =
```
mdedetrich, line 1180: I presume that this type annotation was auto generated from Intellij and if so it likely can be simplified a bit more.

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
```diff
@@ -1177,11 +1177,15 @@ import scala.util.{ Failure, Success, Try }
...
+          SplitAfterSize(chunkSize, chunkBufferSize)(atLeastOneByteString)
+            .via(getChunkBuffer(chunkSize, chunkBufferSize, maxRetries)) // creates the chunks
+
+        val source2 = source1.mergeSubstreamsWithParallelism(parallelism)
```
mdedetrich, line 1184: The `source2` here seems to be a leftover from refactoring/figuring out types locally? You should be able to just go

```scala
source1.mergeSubstreamsWithParallelism(parallelism)
  .filter(_.size > 0)
  .via(atLeastOne)
  .zip(requestInfoOrUploadState(s3Location, contentType, s3Headers, initialUploadState))
  // etc etc
```

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/javadsl/S3.scala
```diff
@@ -485,8 +486,11 @@ object S3 {
    * @return A [[pekko.stream.javadsl.Source]] containing the objects data as a [[pekko.util.ByteString]] along with a materialized value containing the
    *         [[pekko.stream.connectors.s3.ObjectMetadata]]
    */
-  def getObject(bucket: String, key: String): Source[ByteString, CompletionStage[ObjectMetadata]] =
-    new Source(S3Stream.getObject(S3Location(bucket, key), None, None, S3Headers.empty).toCompletionStage())
+  def getObject(bucket: String, key: String): Source[ByteString, CompletionStage[ObjectMetadata]] = {
+    val objectSource: SourceToCompletionStage[ByteString, ObjectMetadata] =
```
mdedetrich, line 490: I think this type can be simplified but as before will have to check out locally (also applies for other future instances).

mdedetrich (changes requested):
There seems to be a case of a leftover refactoring (see comments). Also to me it seems like the `legal/GoogleCommonLicense.txt` is unrelated to this PR???

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
pjfanning: the original code wouldn't compile in scala 3 and splitting the code up has helped to get it to compile

pjfanning:
> There seems to be a case of a leftover refactoring (see comments). Also to me it seems like the `legal/GoogleCommonLicense.txt` is unrelated to this PR???

that change is part of this PR - up until now, the only LICENSE extras were for google-common - so the build was using the root LICENSE for the google-common jar - now we have s3 stuff in the root LICENSE, so I need to create a special license for google-common that omits the s3 stuff

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
pjfanning: code no longer compiles when I start removing the types on these params

mdedetrich: I can see that for `source1`, its `source2` that seems redundant (`source2` has no type annotation).

mdedetrich:
@pjfanning Is this ready or is it still in draft status?

pjfanning:
> @pjfanning Is this ready or is it still in draft status?

The code works but the type annotations to help the Scala 3 compiler are not ideal.

mdedetrich:
> The code works but the type annotations to help the Scala 3 compiler are not ideal.

Ill checkout hte code locally to see what is going on but either way I don't think this is a deal breaker

mdedetrich (approved):
@pjfanning lgtm, I was able to do some fixes/cleanup some other things but I couldn't remove the annoying path dependant types (seems to be another regression with Scala 3 that we need to upstream).

PR #170, opened 2023-06-11
Scala3 ftp support

pjfanning:
part of #126

### DIFF
--- a/ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/javadsl/FtpApi.scala
+++ b/ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/javadsl/FtpApi.scala
@@ -30,7 +30,7 @@ import net.schmizz.sshj.SSHClient
 import org.apache.commons.net.ftp.{ FTPClient, FTPSClient }
 
 @DoNotInherit
-sealed trait FtpApi[FtpClient, S <: RemoteFileSettings] { _: FtpSourceFactory[FtpClient, S] =>
+sealed trait FtpApi[FtpClient, S <: RemoteFileSettings] { self: FtpSourceFactory[FtpClient, S] =>
 
   /**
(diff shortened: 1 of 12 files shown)


ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/javadsl/FtpApi.scala
```diff
@@ -573,6 +573,6 @@ object Sftp extends SftpApi {
    */
   def create(customSshClient: SSHClient): SftpApi =
     new SftpApi {
-      override val sshClient: SSHClient = customSshClient
```
mdedetrich, line 576: Why is this a `def` now? This could have a performance impact because it would get evaluated each time

mdedetrich (changes requested):
Is the change from `def` to `val` necessary?

ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/javadsl/FtpApi.scala
pjfanning: * more scala3 compiler issues
* the super class has `def sshClient(): SSHClient` 
* the `()` seem to be the reason that the override can't make it a `val`
* `customSshClient` is a constant as far as the anonymous class `new SftpApi {` is concerned
   * so there will be no recalculation on each call to `def sshClient(): SSHClient`
   * one option would be to make it `final def sshClient(): SSHClient` -- but with anonymous classes, you would assume everything is implied to be final anyway

mdedetrich: Okay all good then, approving

mdedetrich (approved):
lgtm

Issue #169, opened 2023-06-11
upgrade commons-net to 3.10.0


pjfanning:
Due to https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-37533

We have been wary of updating dependency versions for 1.0.0. The FTP connector is quite self contained and upgrading its dependencies will only impact this one connector.

pjfanning:
#171 was merged

PR #171, opened 2023-06-11
upgrade commons-net to 3.10.0 due to cve

pjfanning:
see #169 

Remaining work
* add config to allow commons-net FTPSClient to be used
* add FTP NOTICE file to connectors-ftp jar
* the LegacyFtpsClient can be greatly simplified after commons-net 3.11.0 is released - we can subclass commons-net FTPSClient and just override one method - https://github.com/apache/commons-net/pull/204 makes the state that we need accessible

### DIFF
--- a/ftp/src/main/java/org/apache/pekko/stream/connectors/ftp/impl/LegacyFtpsClient.java
+++ b/ftp/src/main/java/org/apache/pekko/stream/connectors/ftp/impl/LegacyFtpsClient.java
@@ -1,0 +1,840 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.connectors.ftp.impl;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+import java.net.Socket;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLException;
+import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+
+import org.apache.commons.net.ftp.*;
+import org.apache.commons.net.util.Base64;
+import org.apache.commons.net.util.SSLContextUtils;
+import org.apache.commons.net.util.SSLSocketUtils;
+import org.apache.commons.net.util.TrustManagerUtils;
+
+/**
+ * This class is a workaround for code introduced in commons-net 3.9.0 that breaks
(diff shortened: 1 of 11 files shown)


pjfanning:
so far, this is causing problems with org.apache.pekko.stream.connectors.ftp.FtpsWithProxyStageTest

pjfanning:
test issue possibly caused by changes in https://issues.apache.org/jira/browse/NET-711 or https://issues.apache.org/jira/browse/NET-642

pjfanning:
Since https://issues.apache.org/jira/browse/NET-718 is not making progress - I have brought the old working code from commons-net 3.8.0 into my PR.

Aim is to add a config setting that lets users choose to use the latest commons-net FTPSClient or to use the legacy one. The legacy one will be default since we know that that works with our tests - while the change in commons-net has no tests but theoretically fixes issues for some users (https://issues.apache.org/jira/browse/NET-642)

laglangyue:
lgtm

nvollmar (approved):
lgtm

ftp/src/main/java/org/apache/pekko/stream/connectors/ftp/impl/LegacyFtpsClient.java
```diff
@@ -0,0 +1,836 @@
...
+    /**
+     * Returns a socket of the data connection. Wrapped as an {@link SSLSocket}, which carries out handshake processing.
+     *
+     * @param command The int representation of the FTP command to send.
+     * @param arg     The arguments to the FTP command. If this parameter is set to null, then the command is sent with no arguments.
+     * @return corresponding to the established data connection. Null is returned if an FTP protocol error is reported at any point during the establishment and
+     *         initialization of the connection.
+     * @throws IOException If there is any problem with the connection.
+     * @see FTPClient#_openDataConnection_(int, String)
+     * @deprecated (3.3) Use {@link FTPClient#_openDataConnection_(FTPCmd, String)} instead
```
He-Pin, line 200: Should we adopt the @since version in this file ?

He-Pin (approved):
Lgtm

ftp/src/main/java/org/apache/pekko/stream/connectors/ftp/impl/LegacyFtpsClient.java
pjfanning: The code is not public - ie the class is not public.

Roiocam (approved):
LGTM, I have read about the context of this PR, I have no problem with retaining the legacy client and set as default. But I have a question that will not block this PR, that is, the new `FTPClient` will not work under the proxy settings. Is this limited to testing or general? Should we mention it in the document or the release notes?

ftp/src/main/java/org/apache/pekko/stream/connectors/ftp/impl/LegacyFtpsClient.java
mdedetrich: > This is a temporary copy of the commons-net code. We can remove it when commons-net 3.11.0 is released.

How serious is this cve, can we just wait until commens-net 3.11.0 is released?

If it's that critical I would have assumed that commons-net 3.11.0 would be out by now

pjfanning:
> LGTM, I have read about the context of this PR, I have no problem with retaining the legacy client and set as default. But I have a question that will not block this PR, that is, the new `FTPClient` will not work under the proxy settings. Is this limited to testing or general? Should we mention it in the document or the release notes?

We will include release notes describing the new setting that controls whether to use the legacy or the new FTPS Client.

ftp/src/main/java/org/apache/pekko/stream/connectors/ftp/impl/LegacyFtpsClient.java
mdedetrich: Given this, doesn't it just make sense to wait till 3.11.0 is released rather than prematurely rushing to inline the code now?

pjfanning: The issue is that there is a CVE in the SFTP code of commons-net, fixed in v3.9.0. Unfortunately, commons-net 3.9.0 also includes a change in the FTPS code. The contributor and some other commenters claimed that it fixed FTPS voa proxy for them - without providing evidence or a test case. The equivalent test case in Akka/Pekko (FTPS via proxy fails with the change). So this code copies in the old FTPS code and lets Pekko users choose if they want the old or new FTPS code. I submitted a fix for commons-net 3.11.0 that allows the old code to be shorter becuase it can extend the commons-net FTPSClient. It can't extend the commons-net FTPSClient in commons-net 3.10.0 because too much state is private.

pjfanning:
I logged https://github.com/apache/incubator-pekko-connectors/issues/465 as a follow up

PR #174, opened 2023-06-12
suggested change to queue method call

pjfanning:
based on https://github.com/lampepfl/dotty/issues/17946#issuecomment-1587664760


mdedetrich (approved):
lgtm

PR #179, opened 2023-06-23
upgrade grpc plugin

pjfanning:
the current version that we use has problems causing build failures in pekko samples - https://github.com/apache/incubator-pekko-samples/pull/46


He-Pin (approved):
lgtm

PR #183, opened 2023-06-26
more pekko branding

pjfanning:
* we haven't yet rewritten https://github.com/apache/incubator-pekko-connectors-samples/ but it is better to link here than to akka.io
* with the Kinesis alternatives, I've logged issues in those projects
  * https://github.com/500px/kinesis-stream/issues/15
  * https://github.com/StreetContxt/kcl-akka-stream/issues/191


He-Pin (approved):
lgtm

PR #186, opened 2023-06-26
s3: fix infinite recursion

pjfanning:
function calls itself recursively infinitely - needs to call another overloaded version of the function


mdedetrich (approved):
lgtm

PR #198, opened 2023-07-21
remove code copied from gist

pjfanning:
* the code has no license specified
* if it turns out that we need this test code, we will need to either see if scalatest supports retries now (`eventually` works well)
* or approach the author to see if he will grant us permission to license the gist


mdedetrich:
I think this can be replaced with what I did in https://github.com/apache/incubator-pekko/pull/471 or with something like

```scala
lazy val root = Project("test", file(".")).
  settings(commands += Command.command("testUntilFailed") { state =>
    "test" :: "testUntilFailed" :: state
  })
```

Which lets you run tests until they fail. I am actually thinking of putting both of these into an sbt-plugin (under my org) because they are generally useful.

mdedetrich (approved):
lgtm

PR #206, opened 2023-07-30
Add sbt-license-report

mdedetrich:
Almost done, only some last minute changes need to be done (and maybe link validator checking, although link validator needs to be added in a different repo).


nvollmar (approved):
lgtm

PR #207, opened 2023-07-31
Add commands to apply and check codestyle

mdedetrich:
As can be seen from https://github.com/apache/incubator-pekko-connectors/pull/205#issuecomment-1657823432 users unfamiliar to the codebase are having issues both checking/applying scalafmt and javafmt to the codebase

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -98,6 +98,11 @@ lazy val `pekko-connectors` = project
         |
         |  mimaReportBinaryIssues - checks whether this current API
         |    is binary compatible with the released version
+        |
+        |  checkCodeStyle - checks that the codebase follows code
+        |    style
+        |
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -124,12 +129,8 @@ lazy val `pekko-connectors` = project
     crossScalaVersions := List() // workaround for https://github.com/sbt/sbt/issues/3465
   )
 
-TaskKey[Unit]("verifyCodeFmt") := {
```
mdedetrich, line 127: I have opted to replace this because the current style is overcomplicated especially when compared to `addCommandAlias`. This style made more sense in the past when we weren't checking that code was formatted on PR but this is no longer the case

pjfanning (approved):
lgtm

PR #212, opened 2023-08-06
sbt-source-dist 0.1.10

pjfanning:
switches to just publishing a tgz release (no zip)

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -11,7 +11,7 @@ addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.3")
 addSbtPlugin("com.dwijnand" % "sbt-dynver" % "4.1.1")
 addSbtPlugin("net.bzzt" % "sbt-reproducible-builds" % "0.31")
 addSbtPlugin("org.mdedetrich" % "sbt-apache-sonatype" % "0.1.10")
-addSbtPlugin("com.github.pjfanning" % "sbt-source-dist" % "0.1.8")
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
lgtm

He-Pin:
tgz seems not easy to uncompress on Windows PC.

pjfanning:
> tgz seems not easy to uncompress on Windows PC.

Unfortunately, that is a Windows problem - there must be a tool that you can install on Windows that can untar a file. I haven't used Windows in many many years. I used to use WinZip and it covered all sorts of archive formats.

He-Pin:
Let's follow the ASF rule, haozip can do this or 7zip.

mdedetrich:
> tgz seems not easy to uncompress on Windows PC.

7zip is a program which you can use to decompress tgz files.

I am personally not that pushy when it comes to removing zip support, it was originally suggested as a solution due to a specific test but it turns out the test also has issues with the tgz archive.

He-Pin:
image
@pjfanning Can we provide both?

pjfanning:
> image @pjfanning Can we provide both?

I don't understand how netty4 affects this sbt plugin PR. Is this meant for another PR?

pjfanning:
The discussion was at https://lists.apache.org/thread/s4omnkqxnlsmm7hv4g18q9m4ho5kkdjr and not producing 2 release archives was actually suggested by Ryan Skraba on one of our release votes. The fact that there are some test issues when using the pekko http releases is not the main reason. It saves release effort and voting effort.

He-Pin:
@pjfanning I see.

Issue #151, opened 2023-06-09
scala3 support for Slick connector


pjfanning:
Blocked by https://github.com/slick/slick/issues/2177

KuceraMartin:
Should not be marked as blocked anymore.

pjfanning:
will be in the 1.1.0-M1 release

PR #223, opened 2023-08-14
scala3 support for slick

pjfanning:
* relates to #151 
* I don't intend to merge this until Slick 3.5.0 is released


mdedetrich:
So the slick tests are definitely flaky, I added a commit which seems to have helped (i.e. only failing ~30% of the time). I don't recall if the Slick tests were normally this flaky (don't remember this being the case) and it only seems to be happening on Scala 3.

pjfanning:
The Slick tests haven't been all that flaky up until now.

mdedetrich:
> The Slick tests haven't been all that flaky up until now.

Yes this is my recollection as well, the flakiness is only occurring with Slick Scala 3

mdedetrich:
@pjfanning There are MiMa related failures, although tbh we never agreed for MiMa on this pekko module and Slick 3.5 is slated for 1.1.x anyways

mdedetrich:
@pjfanning I just cleaned up the branch which included some rebases so I would just pull the latest changes ignoring the local ones.

Good news is that Slick 3.5.0-M4 seems to work fine with none of the previous problems that we were experiencing.

mdedetrich:
So it is being [asked](https://github.com/slick/slick/discussions/2891#discussioncomment-8550269) if we can publish these changes so that people can test Slick 3.5.0-RC1 before Slick does a full release. Since 1.0.x branch is already setup if we want to do snapshots we can just merge this PR as is, milestone would obviously require a proper vote which I am not against and given what we discussed at https://github.com/apache/incubator-pekko-connectors/discussions/443#discussioncomment-8506668 creating a milestone right now (aside from just Slick 3.5.0-RC1) may not be a bad idea since we already did quite a few major dependency updates.

@pjfanning wdyt?

raboof:
Ideally I think releases should not depend on snapshot/RC dependencies, and the main branch should always be in a release-able state. I guess we can deviate from that if there is a strong reason for it. Alternatively, would it be possible to set up our GitHub Actions so that we can publish a snapshot built off a branch/PR?

pjfanning:
I agree that it is too time consuming for us to do releases to test non-ASF jars. If someone really wanted to test this, they could build pekko-connectors-slick jars themselves by doing a git checkout of this PR's branch and running sbt commands.

mdedetrich:
> Alternatively, would it be possible to set up our GitHub Actions so that we can publish a snapshot built off a branch/PR?

I wanted to do this at some point I just never got around to it. Note that nothing is stopping us from publishing a snapshot locally from your machine, its just that you have to be careful about the version name that you set as it can break https://github.com/pjfanning/sbt-pekko-build/blob/main/src/main/scala/com/github/pjfanning/pekkobuild/PekkoDependency.scala

mdedetrich:
Replied at https://github.com/slick/slick/discussions/2891#discussioncomment-8554394

slick/src/main/scala/org/apache/pekko/stream/connectors/slick/javadsl/Slick.scala
```diff
@@ -54,6 +51,7 @@ object Slick {
       session: SlickSession,
       query: String,
       mapper: JFunction[SlickRow, T]): Source[T, NotUsed] = {
+
```
mdedetrich, line 54: nit: Random new line?

slick/src/test/java/docs/javadsl/SlickTest.java
mdedetrich, line 1: Can you try reverting the change in this file from `10` seconds to `5` seconds? This was done in diagnosing an error with flaky tests on an earlier Slick version however it was unrelated (and latest slick version no longer has this flakiness)

mdedetrich (changes requested):
Just one change I would like to explore

slick/src/main/scala/org/apache/pekko/stream/connectors/slick/javadsl/Slick.scala
pjfanning: just spreading out the code to make it more readable

mdedetrich (approved):
lgtm, CI still passes even with test timeouts reverted

PR #227, opened 2023-08-19
Add scala3 support to Azure Storage Queue

mdedetrich:
Either this may have been missed or we are lucky because the azure-storage-queue compiles with Scala 3 fine


pjfanning (approved):
lgtm

PR #231, opened 2023-08-20
grpc 1.0.0 release

pjfanning:
we still need to have the snapshot resolver because of the pekko-sbt-paradox plugin and pekko-material-theme


mdedetrich (approved):
lgtm

PR #252, opened 2023-09-20
Fix apiURL for autoAPIMappings

mdedetrich:
Pekko projects use a feature called `autoAPIMappings` which allows a project itself to specify where its docs are published to so that when other projects include this project as a dependency, scaladoc/javadoc automatically know. where to link the URL.

This is was the core issue behind https://github.com/apache/incubator-pekko-persistence-cassandra/pull/102#issue-1899886190.

@pjfanning This may be strong enough grounds for a new release because basically **any** project that references anything in Pekko Connectors will have incorrect links in docs.


project/Common.scala
```diff
@@ -111,8 +111,10 @@ object Common extends AutoPlugin {
       case _ => Seq.empty[String]
     }),
     autoAPIMappings := true,
-    apiURL := Some(url(
-      s"https://pekko.apache.org/api/pekko-connectors/${version.value}/org/apache/pekko/stream/connectors/index.html")),
+    apiURL := {
```
mdedetrich, line 114: Taken from https://github.com/apache/incubator-pekko-http/blob/75f26d6c45d4f0ea78737d8daf96020b3327f703/build.sbt#L37-L40 so we know it should work.

He-Pin (approved):
lgtm

PR #256, opened 2023-10-07
upgrade google-auth dependency

pjfanning:
google-auth-library-oauth2-http uses an old version of Jackson that has CVEs


mdedetrich (approved):
lgtm

PR #260, opened 2023-10-09
Remove unnecessary license comments

mdedetrich:
I think this was done before but because of  issues when rebasing the scala3 branch the changes were lost. In any case, these comments are unnecessary because sbt-license-report does the same job (and more correctly since it actually scans the dependency tree).


pjfanning (approved):
lgtm

PR #261, opened 2023-10-11
upgrade aws libs

pjfanning:
* may not be that useful - I thought aws sdk used netty but it seems to just use it in test cases
* still may be useful to upgrade but maybe only in pekko-connectors 1.1.0


pjfanning:
What does everyone think? Safe for the connectors v1.0.1 release or hold it back for the v1.1.0 release?

mdedetrich:
To me its safer to do in 1.1.x especially considering that we already did a 1.0.0 release

samueleresca (approved):

PR #263, opened 2023-10-16
upgrade jackson version

pjfanning:
* try to get orientdb and cassandra connectors to use jackson 2.14.3
* also some jackson modules in geode are behind
* jackson 2.14.3 is what we use in pekko and it has no CVEs


mdedetrich (approved):
lgtm as long as all tests pass

PR #276, opened 2023-11-01
Remove useless condition

mdedetrich:
image


He-Pin:
Maybe the code is leaves for easy understanding ？

He-Pin (approved):
Lgtm

mdedetrich:
> Maybe the code is leaves for easy understanding ？

Actually I thin its a mistake because it was manually translated from https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.client.mqttv3/src/main/java/org/eclipse/paho/client/mqttv3/MqttTopic.java#L253-L319

The main thing to be careful is that whether this inspection is correct (from first glance this does appear to be the case).

mdedetrich:
Let me merge this PR myself, since we are dealing with mqtt I need to re-run the tests until the pass (just to make sure) and I will also wait for one more reviewer just to be safe.

mdedetrich:
Okay all tests pass, lets wait for another review incase any one of us missed something.

nvollmar (approved):
lgtm

PR #279, opened 2023-11-12
Accept any materializer type param for S3's chunkUploadSink

mdedetrich:
When I originally added this functionality I mistakingly hardcoded the type parameter to `NotUsed` rather than just using `_` (which means don't care). While this may initially seem like a non-issue, a problem eventuates if you happen to use a `Sink` that materializes a value that is contained within a `Future` i.e. `Sink[(UploadPartResponse, immutable.Iterable[C]), Future[Done]]`. Such sinks are quite common and there is no way to get a value outside of a `Future` aside from blocking (which you should never do in production code) in order to satisfy `chunkUploadSink: Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed]`.

Note that since `chunkUploadSink` ultimately ends up [getting used in `.alsoTo`](https://github.com/apache/incubator-pekko-connectors/blob/18fc2932cf78194ea149bda0d9e73f7161b8e39f/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala#L1323) whos materializer type argument param is also `_` (see https://github.com/apache/incubator-pekko/blob/1b1f57224b409c8b2cbd5267ace814439e97d3ea/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala#L3464) and the fact that we are only dealing with type parameters which get erased at runtime means that this change is completely safe for the 1.0.x series (in fact it shouldn't even generate any bytecode difference either way, all this does is make scalac satisfy code where the materializer type param is not `NotUsed`)

### DIFF
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
@@ -906,7 +906,7 @@ import scala.util.{ Failure, Success, Try }
    */
   def multipartUploadWithContext[C](
       s3Location: S3Location,
-      chunkUploadSink: Sink[(UploadPartResponse, immutable.Iterable[C]), NotUsed],
+      chunkUploadSink: Sink[(UploadPartResponse, immutable.Iterable[C]), _],
       contentType: ContentType = ContentTypes.`application/octet-stream`,
       s3Headers: S3Headers,
       chunkSize: Int = MinChunkSize,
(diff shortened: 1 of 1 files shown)


pjfanning (commented):
We don't have mima checks enabled on this repo. Could you merge that PR and rebase to see if this change needs a mima exclusion?

mdedetrich:
Sure

mdedetrich:
@pjfanning https://github.com/apache/incubator-pekko-connectors/pull/278 has been approved

pjfanning:
> @pjfanning #278 has been approved

That PR is merged. Could you rebase this?

mdedetrich:
@pjfanning The relevant tests passed (including MiMa check)

pjfanning (approved):
Lgtm

PR #284, opened 2023-11-24
make dependence on jackson-databind explicit (bigquery connector)

pjfanning:
BigQuery connector has jackson code and it messily only updates some jackson jar dependencies while leaving some key ones not updated.


pjfanning:
Probably best left alone - the pekko-http-jackson dependency is `provided` so it is probably better not to add an explicit dependency on jackson-databind

PR #288, opened 2023-12-02
Update Pekko

mdedetrich:
Updates Pekko to 1.0.2

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -20,7 +20,7 @@ object Dependencies {
   val Scala3 = "3.3.1"
   val ScalaVersions = Seq(Scala213, Scala212, Scala3)
 
-  val PekkoVersion = "1.0.1"
+  val PekkoVersion = "1.0.2"
   val PekkoBinaryVersion = "1.0"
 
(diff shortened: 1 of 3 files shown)


mdedetrich:
@pjfanning Was there some change in Pekko 1.0.2 that would cause pekko-connectors to fail to compile, specifically for the xml module?

pjfanning:
> @pjfanning Was there some change in Pekko 1.0.2 that would cause pekko-connectors to fail to compile, specifically for the xml module?

not sure why this is happening but the logs don't highlight a compile issue, maybe there is a compile warning now and the build is set up to fail with compile warnings?

mdedetrich:
> not sure why this is happening but the logs don't highlight a compile issue, maybe there is a compile warning now and the build is set up to fail with compile warnings?

Let me confirm this then, it's suspicious that the build fail didn't occur in the PR that changed the compile warnings.

mdedetrich:
Okay so the nowarning's did fix the issue, everything is good now

pjfanning (approved):
lgtm

PR #289, opened 2023-12-09
logback 1.2.11

pjfanning:
cassandra tests timeout with logback 2.13


pjfanning:
these tests are still failing - looks like it is coincidental that they started failing around the time of the logback upgrade

PR #290, opened 2023-12-09
increase cassandra test timeout

pjfanning:
Cassandra tests started failing recently and it isn't clear that any recent change is causing this. I tried reverting the logback upgrade but that didn't help.

This timeout change seems to help.


mdedetrich (approved):
lgtm

PR #295, opened 2023-12-23
prepend package name for Record class

pjfanning:
clashes with java.lang.Record

causes compile issues if you build with JDK >= 14

### DIFF
--- a/.github/workflows/nightly-pekko-1.1-builds.yaml
+++ b/.github/workflows/nightly-pekko-1.1-builds.yaml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


mdedetrich (approved):
lgtm although its cleaner to just use an import alias i.e.

```scala
import software.amazon.awssdk.services.kinesis.model.{Record => KRecord}
```

pjfanning:
> lgtm although its cleaner to just use an import alias i.e.
> 
> ```scala
> import software.amazon.awssdk.services.kinesis.model.{Record => KRecord}
> ```

this is a Java file though

mdedetrich:
> this is a Java file though

Woops missed this, nw then

PR #298, opened 2023-12-27
Kinesis: fix maxBytesPerSecond throttling

jtjeferreira:
The kinesis throttling is not being respected because the function `getPayloadByteSize` is incorrect. 

```scala
private def getPayloadByteSize[T](record: (PutRecordsRequestEntry, T)): Int = record match {
    case (request, _) => request.partitionKey.length + request.data.asByteBuffer.position()
}
```

`request.data.asByteBuffer.position()` can return 0

The ticket was originally reported in https://github.com/akka/alpakka/issues/2871 and I did the fix in the akka repository. AFAICT I can use the same code as long as I am the author...


pjfanning (approved):
Lgtm - thanks

mdedetrich (approved):
lgtm. If we wan't to be strict this should land in 1.1.0 since this is a behaviour change but afaict the current functionality appears to be broken so this can be an exception

@jtjeferreira Can you confirm?

jtjeferreira:
> lgtm. If we wan't to be strict this should land in 1.1.0 since this is a behaviour change but afaict the current functionality appears to be broken so this can be an exception
> 
> @jtjeferreira Can you confirm?

yes the current functionality is broken

pjfanning:
Merging - as it is a bug fix

PR #299, opened 2023-12-29
Remove redundant formatting sbt settings

mdedetrich:
We have long since removed the concept of formatting sources on compile in all of the Pekko modules, furthermore setting these settings are redundant anyways since they default to `false` anyways.


pjfanning (approved):
lgtm

PR #301, opened 2023-12-30
[Draft]fix slick java test 

laglangyue:
todo

### DIFF
--- a/slick/src/test/java/docs/javadsl/DocSnippetFlow.java
+++ b/slick/src/test/java/docs/javadsl/DocSnippetFlow.java
@@ -17,6 +17,7 @@ import org.apache.pekko.Done;
 import org.apache.pekko.actor.ActorSystem;
 import org.apache.pekko.stream.connectors.slick.javadsl.Slick;
 import org.apache.pekko.stream.connectors.slick.javadsl.SlickSession;
+import org.apache.pekko.stream.connectors.slick.javadsl.SlickSession$;
 import org.apache.pekko.stream.javadsl.Sink;
 import org.apache.pekko.stream.javadsl.Source;
 
@@ -31,7 +32,7 @@ public class DocSnippetFlow {
     final ActorSystem system = ActorSystem.create();
 
     // #flow-example
-    final SlickSession session = SlickSession.forConfig("slick-h2");
+    final SlickSession session = SlickSession$.MODULE$.forConfig("slick-h2");
     system.registerOnTermination(session::close);
 
     final List<User> users =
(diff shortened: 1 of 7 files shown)


laglangyue:
Could you please open CI? There seems to be some issues with using scala objects in `docs.javadsl`
@pjfanning

laglangyue:
this pr includes some `clean code`,I will revert later.

pjfanning:
I would prefer not to run this. You can experiment on your own fork. Or you can test on your laptop.

To test on your fork, enable CI in setup. And you can modify the workflow yml to remove the condition that enforces that the workflow only runs for the Apache fork of this repo.

pjfanning:
I will not merge any PRs that do not fully explain the reason for the changed. Todo is not an acceptable description. I have no idea why the changed in this PR are needed.

slick/src/test/java/docs/javadsl/DocSnippetFlow.java
```diff
@@ -31,7 +32,7 @@ public static void main(String[] args) throws Exception {
     final ActorSystem system = ActorSystem.create();
 
     // #flow-example
-    final SlickSession session = SlickSession.forConfig("slick-h2");
+    final SlickSession session = SlickSession$.MODULE$.forConfig("slick-h2");
```
pjfanning, line 35: this is not needed and is less readable - I tested that the existing code compiles

```
sbt slick/test
```

I deliberately changed the Java code to add an unexpected character and the sbt command above failed - meaning that that command does compile the Java tests.

laglangyue: My idea will report an error. 
image

It maybe a bug for idea.In fact, I also test passed locally, then I wan to know how ci work with it, so I open this PR,because of lacking of github action.
I will report this for idea-scala-plugin.

laglangyue: I understand it,it truely a bug for idea, because the bytecode is changed.
image

pjfanning: GitHub CI only runs automatically for members of the Apache GitHub group. You will have test this stuff on your laptop.

IDEA has lots of issues with compiling Scala. It gets even worse if you want to compile Scala 3.

laglangyue:
```scala
class TestObject

class TestHello {
  def hello() = {
    println("ok")
  }
}

object TestObject extends TestHello
```
image

I write a min code of this.
Idea will not prompt  methods of object which is extended.

Issue #304, opened 2024-01-02
The used version of sshj is vulnerable to CVE-2023-48795


DarkDrek:
https://github.com/apache/incubator-pekko-connectors/blob/0ef66493cbebf792aa675e30cabc1e7d81e5e964/project/Dependencies.scala#L182

The ftp connector uses sshj in version 0.33.0 which is vulnerable to https://nvd.nist.gov/vuln/detail/CVE-2023-48795 
It should be updated to at least 0.38.0 see warning in https://github.com/hierynomus/sshj

pjfanning:
Thanks - we intend to upgrade some dependencies for the 1.1.0 release. You should be able to upgrade your own dependencies in your build script - you don't need to rely on us to release jars built with the latest and greatest transitive dependency releases.

pjfanning:
PS we will set up scala steward so we will get automated creation of PRs created for dependency upgrades.

pjfanning:
https://github.com/hierynomus/sshj release notes mention the CVE

pjfanning:
#305 merged and it will be in v1.0.2 which we hope to release in the next few weeks.

DarkDrek:
Thanks for the fast work.

PR #305, opened 2024-01-02
sshj 0.38.0

pjfanning:
see #304

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -41,8 +41,9 @@ object Dependencies {
   val hoverflyVersion = "0.14.1"
   val scalaCheckVersion = "1.16.0"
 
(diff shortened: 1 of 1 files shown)


pjfanning:
@mdedetrich @kerr what do you think about the idea of including this in the pekko-connectors 1.0.2 release since it has a security impact? See https://github.com/apache/incubator-pekko-connectors/issues/304

mdedetrich (approved):
lgtm. I am fine with merging it for pekko-connectors 1.0.2 however would also like to get @He-Pin feedback

He-Pin:
Supply chain security is always a concern,  and as we don't have a 1.1.x, we should fixed it in 1.0.x then.

He-Pin (approved):
Lgtm

PR #306, opened 2024-01-02
[slick][javadsl] improve SlickSession.forConfig usage with idea

laglangyue:
## motivation
If the Companion object `A` of `class A` extend  a class `B`, the idea will report an error, because idea can't find the method include the class `B`

image
user need to use `SlickSession$.MODULE$.forConfig("slick-h2");`, this usage connot match the description on the official website.

Report it to idea,but idea may not be fixed or slowly, as they are mainly Java ideas and cannot fix such issues. Previously reported an error with the inline.

## how to do
just wrapper the extend method

## test 
had test locally.

### DIFF
--- a/slick/src/main/scala/org/apache/pekko/stream/connectors/slick/javadsl/package.scala
+++ b/slick/src/main/scala/org/apache/pekko/stream/connectors/slick/javadsl/package.scala
@@ -59,7 +59,12 @@ private[slick] abstract class SlickSessionFactory {
  * closed after creation to avoid leaking database resources like active
  * connection pools, etc.
  */
-object SlickSession extends SlickSessionFactory
+object SlickSession extends SlickSessionFactory {
+  override def forConfig(path: String) = super.forConfig(path)
+  override def forConfig(config: Config) = super.forConfig(config)
+  override def forConfig(path: String, config: Config) = super.forConfig(path, config)
+  override def forConfig(databaseConfig: DatabaseConfig[JdbcProfile]) = super.forConfig(databaseConfig)
+}
 
 /**
  * Java API: A class representing a slick resultset row, which is used


pjfanning:
Let's wait for IDEA to get back. I don't believe it is up to us to hack our code due to issues in IDEA.

laglangyue:
> Let's wait for IDEA to get back. I don't believe it is up to us to hack our code due to issues in IDEA.

I think this is only an issue with the idea, and it is also a lack of comprehensive consideration by the slide API designer, without good abstraction. 
Java DSL and Scala DSL are not asymmetric，and api should be included in  `Companion object`, just use Object.

The perfect Java API implementation should be 
pekko-connector-cassandraSource
https://github.com/apache/incubator-pekko-connectors/blob/main/cassandra/src/main/scala/org/apache/pekko/stream/connectors/cassandra/javadsl/CassandraSource.scala
spark
https://github.com/apache/spark/blob/master/sql/api/src/main/java/org/apache/spark/sql/types/DataTypes.java

laglangyue:
https://youtrack.jetbrains.com/issue/SCL-12670/IntelliJ-cannot-resolve-methods-from-companion-objects-of-sealed-traits-in-an-imported-library
this bug has existed 4 years

laglangyue:
> https://github.com/slick/slick is a designed for Scala users. Is it even that usable by Java users? Wouldn't you be better off with something like Hibernate?

Indeed, I don't use JavaDSL either, so I didn't make any major changes. But we provide JavaDSL.

laglangyue:
https://youtrack.jetbrains.com/issue/SCL-21966/Cannot-resolve-method-which-extends-from-class-for-a-Companion-object
I has reported it to idea scala.

mdedetrich (changes requested):
I am going to block this PR in agreement with @pjfanning , we shouldn't be making significant changes to the codebase just to please an editor/IDE.

Not everyone uses Intellij and there is an implicit understanding that when someone reads code, its because its relevant to the problem being solved and not pleasing the editor/IDE.

Best course of action is to file a bug report which appears to already be done.

PR #311, opened 2024-01-09
support ftps implicit mode

pjfanning:
* one issue is that commons-net FTPSClient needs the implicit flag set when it is constructed and this seems to happen before we even have FtpsSettings available to look at - it could took a bit of refactoring to get the value from FtpsSettings
* some package private code was changed so I wouldn't regard this as something that should be backported to 1.0.x
* the public API is unaffected


laglangyue:
Overall, LGTM, and I leave some opinion here
(1) Is this compatible with historical users?
(2) should we update the doc `docs/src/main/paradox/ftp.md` ?

pjfanning:
> Overall, LGTM, and I leave some opinion here (1) Is this compatible with historical users? (2) should we update the doc `docs/src/main/paradox/ftp.md` ?

the public API is unaffected - nothing in ftp.md needs to change

build.sbt
```diff
@@ -13,6 +13,7 @@ ThisBuild / apacheSonatypeProjectProfile := "pekko"
 sourceDistName := "apache-pekko-connectors"
 sourceDistIncubating := true
 ThisBuild / resolvers += Resolver.ApacheMavenSnapshotsRepo
+ThisBuild / resolvers += Resolver.ApacheMavenStagingRepo
```
mdedetrich, line 16: I think this is left over from an earlier point in time, we rebase against`main`?

pjfanning: removed and rebased

ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/model.scala
```diff
@@ -247,6 +252,7 @@ final class FtpsSettings private (
     s"credentials=$credentials," +
     s"binary=$binary," +
     s"passiveMode=$passiveMode," +
+    s"useFtpsImplicit=$useFtpsImplicit," +
     s"autodetectUTF8=$autodetectUTF8" +
```
He-Pin, line 256: "autodetectUTF8=$autodetectUTF8" should be
"autodetectUTF8=$autodetectUTF8,"

He-Pin (approved):
lgtm

ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/model.scala
pjfanning: a pre-existing issue - but I've pushed a fix to this PR

samueleresca (approved):
looks good to me

PR #313, opened 2024-01-14
Use scope config for compute-engine auth

mdedetrich:
Currently when getting Google credentials using the `ComputeEngineCredentials` method, it ignores the [`scopes`](https://github.com/apache/incubator-pekko-connectors/blob/1fe3474d654bf48b0e81011d878e541a32409c93/google-common/src/main/resources/reference.conf#L9) variable.

In the official Google SDK, you can see that it also has the same logic for getting the credentials from `ComputeEngine`, see https://github.com/googleapis/google-auth-library-java/blob/a7a8d7a4102b0b7c1b83791947ccb662f060eca7/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L193-L199 and https://github.com/googleapis/google-auth-library-java/blob/a7a8d7a4102b0b7c1b83791947ccb662f060eca7/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L480-L483 (which is the same URI as https://github.com/apache/incubator-pekko-connectors/blob/67eea4f1dc7e9e0931322ef5c07f305e995b0eb2/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/GoogleComputeMetadata.scala#L33-L34)

### DIFF
--- a/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/ComputeEngineCredentials.scala
+++ b/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/ComputeEngineCredentials.scala
@@ -25,18 +25,19 @@ import scala.concurrent.Future
 @InternalApi
 private[auth] object ComputeEngineCredentials {
 
-  def apply()(implicit system: ClassicActorSystemProvider): Future[Credentials] =
+  def apply(scopes: Set[String])(implicit system: ClassicActorSystemProvider): Future[Credentials] = {
     GoogleComputeMetadata
       .getProjectId()
-      .map(new ComputeEngineCredentials(_))(system.classicSystem.dispatcher)
+      .map(projectId => new ComputeEngineCredentials(projectId, scopes))(system.classicSystem.dispatcher)
+  }
 
 }
 
 @InternalApi
-private final class ComputeEngineCredentials(projectId: String)(implicit mat: Materializer)
+private final class ComputeEngineCredentials(projectId: String, scopes: Set[String])(implicit mat: Materializer)
     extends OAuth2Credentials(projectId) {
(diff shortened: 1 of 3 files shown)


pjfanning:
you'll need something like https://github.com/apache/incubator-pekko-connectors/pull/311/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R16 to get the mima check to work - v1.0.2-RC1 is current latest tag and this is only in staging repo - this PR looks like it could have a small but non-zero mima footprint

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/ComputeEngineCredentials.scala
```diff
@@ -25,18 +25,19 @@ import scala.concurrent.Future
 @InternalApi
 private[auth] object ComputeEngineCredentials {
 
-  def apply()(implicit system: ClassicActorSystemProvider): Future[Credentials] =
+  def apply(scopes: Set[String])(implicit system: ClassicActorSystemProvider): Future[Credentials] = {
```
He-Pin, line 28: Add a overloaded apply to keep mima happy

mdedetrich: This is internal API and its `private`. MiMa is failing for other reasons (see https://github.com/apache/incubator-pekko-connectors/pull/314#issuecomment-1890890839)

mdedetrich:
@pjfanning @He-Pin You mind re-reviewing this?

nvollmar (approved):
lgtm

PR #314, opened 2024-01-14
Remove duplicate scopes value in reference.conf

mdedetrich:
While working on https://github.com/apache/incubator-pekko-connectors/pull/313 I realized that we have duplicate scope configuration settings, i.e. one [here](https://github.com/apache/incubator-pekko-connectors/blob/1fe3474d654bf48b0e81011d878e541a32409c93/google-common/src/main/resources/reference.conf#L9) and another [here](https://github.com/apache/incubator-pekko-connectors/blob/1fe3474d654bf48b0e81011d878e541a32409c93/google-common/src/main/resources/reference.conf#L31-L32).

This PR removes the scopes config [here](https://github.com/apache/incubator-pekko-connectors/blob/1fe3474d654bf48b0e81011d878e541a32409c93/google-common/src/main/resources/reference.conf#L31-L32) and instead makes the `ServiceAccountCredentials` use the scopes value from [here](https://github.com/apache/incubator-pekko-connectors/blob/1fe3474d654bf48b0e81011d878e541a32409c93/google-common/src/main/resources/reference.conf#L9) which is currently unused.

We also change the collection from a `Seq`/`List` to a `Set` since we want duplicates to be removed and ordering does not matter (also to make it consistent with https://github.com/apache/incubator-pekko-connectors/pull/313)

### DIFF
--- a/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/GoogleOAuth2.scala
+++ b/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/GoogleOAuth2.scala
@@ -37,7 +37,7 @@ private[auth] object GoogleOAuth2 {
 
   private val oAuthTokenUrl = "https://oauth2.googleapis.com/token"
 
-  def getAccessToken(clientEmail: String, privateKey: String, scopes: Seq[String])(
+  def getAccessToken(clientEmail: String, privateKey: String, scopes: Set[String])(
       implicit mat: Materializer,
       settings: RequestSettings,
       clock: Clock): Future[AccessToken] = {
@@ -58,7 +58,7 @@ private[auth] object GoogleOAuth2 {
     }
   }
 
-  private def generateJwt(clientEmail: String, privateKey: String, scopes: Seq[String])(
+  private def generateJwt(clientEmail: String, privateKey: String, scopes: Set[String])(
       implicit clock: Clock): String = {
     import spray.json._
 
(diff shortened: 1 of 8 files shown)


mdedetrich:
Pinging @armanbilge since he originally wrote this, not sure if you wanted to give your 2c

pjfanning:
you'll need something like https://github.com/apache/incubator-pekko-connectors/pull/311/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R16 to get the mima check to work - v1.0.2-RC1 is current latest tag and this is only in staging repo - this PR looks like it could have a big mima footprint

mdedetrich:
This PR doesn't have any MiMa impact, ill just rebase when 1.0.2 is released

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/GoogleOAuth2.scala
```diff
@@ -37,7 +37,7 @@ private[auth] object GoogleOAuth2 {
 
   private val oAuthTokenUrl = "https://oauth2.googleapis.com/token"
 
-  def getAccessToken(clientEmail: String, privateKey: String, scopes: Seq[String])(
+  def getAccessToken(clientEmail: String, privateKey: String, scopes: Set[String])(
```
He-Pin, line 40: I think a mina filter will be needed. or add an overloaded method which use seq.toSet then?

mdedetrich: I'll wait till 1.0.2 is out since that will fix the existing MiMa issue and then I will see.if this is still problematic

jxnu-liguobin (approved):
lgtm

He-Pin (approved):
Lgtm

mdedetrich:
Merging this since `1.0.x` branch exists

pjfanning:
@JD557 I added your note from #861

PR #316, opened 2024-01-16
Add PekkoHttpDependency from sbt-pekko-build

mdedetrich:
Just like with pekko core, adds a dynamic dependency for pekko-http to pekko-connectors so we make sure that pekko-connectors continue to work against new versions of pekko-http (connectors uses pekko-http heavily so we should be doing this).

### DIFF
--- a/project/PekkoHttpDependency.scala
+++ b/project/PekkoHttpDependency.scala
@@ -1,0 +1,24 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import com.github.pjfanning.pekkobuild.PekkoDependency
+
+object PekkoHttpDependency extends PekkoDependency {
+  override val checkProject: String = "pekko-http-testkit"
+  override val module: Option[String] = Some("http")
+  override val currentVersion: String = "1.0.0"
+}
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -31,7 +31,7 @@ object Dependencies {
   val NettyVersion = "4.1.104.Final"
   // Sync with plugins.sbt
   val PekkoGrpcBinaryVersion = "1.0"
-  val PekkoHttpVersion = "1.0.0"
+  val PekkoHttpVersion = PekkoHttpDependency.version
   val PekkoHttpBinaryVersion = "1.0"
   val ScalaTestVersion = "3.2.14"
   val TestContainersScalaTestVersion = "0.40.14"


project/PekkoHttpDependency.scala
```diff
@@ -0,0 +1,7 @@
+import com.github.pjfanning.pekkobuild.PekkoDependency
```
laglangyue, line 1: it seems to lack of license header

laglangyue:
After supplementing the necessnary license header, LGTM

mdedetrich:
> After supplementing the necessnary license header, LGTM

Thanks fixed

mdedetrich:
PR is failing because `-RC1` is not out yet, need to wait for the Pekko Connectors 1.0.2 release to complete

pjfanning (approved):
I think it is quite possible that many connectors will use pekko-http transitively instead of explicitly (via pekko-grpc, aws-spi-pekko-http, etc.) So we might want to add explicit dependencies on pekko-http in these connectors - so that when we run tests, those tests will then use the PekkoHttpDependency version.

mdedetrich:
> I think it is quite possible that many connectors will use pekko-http transitively instead of explicitly (via pekko-grpc, aws-spi-pekko-http, etc.) So we might want to add explicit dependencies on pekko-http in these connectors - so that when we run tests, those tests will then use the PekkoHttpDependency version.

This should already be the case? The dependency resolution should pick the latest explicit version which in this case will be `PekkoHttpDependency.version`.

pjfanning:
> > I think it is quite possible that many connectors will use pekko-http transitively instead of explicitly (via pekko-grpc, aws-spi-pekko-http, etc.) So we might want to add explicit dependencies on pekko-http in these connectors - so that when we run tests, those tests will then use the PekkoHttpDependency version.
> 
> This should already be the case? The dependency resolution should pick the latest explicit version which in this case will be `PekkoHttpDependency.version`.

no it won't - not unless there is an explicit dependency in Dependencies.scala - other it will pick up the pekko-http dependency that is in the pom file of pekko-grpc or other pekko-http dependent dependency.

pjfanning (changes requested):
Could you change PekkoHttpBinaryVersion value? Not to hardcode 1.0.

Also in build.sbt (near bottom of file), change to:
```
  log.info(
    s"Building Pekko Connectors $v against Pekko ${Dependencies.PekkoVersion} and Pekko HTTP ${Dependencies.PekkoHttpVersion} on Scala ${(googleCommon / scalaVersion).value}")
```

mdedetrich:
> no it won't - not unless there is an explicit dependency in Dependencies.scala - other it will pick up the pekko-http dependency that is in the pom file of pekko-grpc or other pekko-http dependent dependency.

Yes but that is the case https://github.com/mdedetrich/incubator-pekko-connectors/blob/99744c0bfbe2680a949b5d79cc5635cf2d02c938/project/Dependencies.scala#L34 is being used as an explicit dependency all throughout pekko connectors.

> Could you change PekkoHttpBinaryVersion value? Not to hardcode 1.0.

Im doing a PR against sbt-pekko-build to add `binaryVersion` to `PekkoDependency`

pjfanning (approved):
lgtm - @mdedetrich could we get this merged so that we can set up 1.0.x branch?

mdedetrich:
@pjfanning So I have a biggish change which I haven't pushed yet to sbt-pekko-build which includes logic for `binaryVersion` which is something that you asked for but I guess that can wait?

pjfanning:
> @pjfanning So I have a biggish change which I haven't pushed yet to sbt-pekko-build which includes logic for `binaryVersion` which is something that you asked for but I guess that can wait?

Can we add that when it's ready and merge this as is?

mdedetrich:
> Can we add that when it's ready and merge this as is?

Fine with me! Was just confirming.

mdedetrich:
@pjfanning So the PR is still failing because Pekko Connectors 1.0.2 is not released yet, this is likely to happen tonight so I will re-run the CI when its pushed and then we can merge.

pjfanning:
it will work if you rebase - I have a PR that depends on this - I will rework mine not to need this

pjfanning:
the mima issue is fixed on main already - thus the usefulness of the rebase

mdedetrich:
> the mima issue is fixed on main already - thus the usefulness of the rebase

Rebased

PR #317, opened 2024-01-20
add warning to init.sh

pjfanning:
@Claudenw requested that we document this file.

https://lists.apache.org/thread/7p1k49ytrwknbcyjwnj0jo1o00z4hnlw

This file is used in our automated tests. We use docker-compose to start an FTP server and a Squid Proxy server.

https://github.com/apache/incubator-pekko-connectors/blob/a639063ffc41a9cb8f9ba5c11dfd997a6750b3c1/docker-compose.yml#L235

### DIFF
--- a/ftp/src/test/resources/sftpd/init.sh
+++ b/ftp/src/test/resources/sftpd/init.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# This script is only meant to be run as part of Docker container.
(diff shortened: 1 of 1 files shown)


pjfanning:
This is mergeable - the build failure is because mima check is broken because the 1.0.2-RC1 is not released - only in staging.

samueleresca (approved):
looks good to me

PR #319, opened 2024-01-20
fix mima check

pjfanning:
broken because 1.0.2-RC1 is only in staging - this issue is probably going to affect us during every release cycle


samueleresca (approved):
looks good to me

PR #327, opened 2024-01-26
Create nested-sample.tar.README

pjfanning:
based on https://github.com/apache/incubator-pekko-http/blob/213a460a2c196617370309216f54803e408d65a5/http-tests/src/test/extra-libs.zip.README

ASF reviewers find our binary files and we need to be explain why we have them checked in.


mdedetrich (approved):
lgtm

PR #329, opened 2024-01-27
chore: Make use of statefulMap instead.

He-Pin:
Motivation:

Make use of statefulMap instead.


google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
```diff
@@ -61,14 +61,9 @@ private[connectors] object ResumableUpload {
...
-            var cumulativeLength = 0L
-            bytes => {
-              val chunk = Chunk(bytes, cumulativeLength)
-              cumulativeLength += bytes.length
-              chunk :: Nil
-            }
-          }
+          .statefulMap(() => 0L)((cumulativeLength, bytes) =>
+              (cumulativeLength + bytes.length, Chunk(bytes, cumulativeLength)),
+            _ => None)
```
He-Pin, line 66: These three flow can be optimized with a single statefulmap I think ,but leave that for another pr.

PR #325, opened 2024-01-25
protect 1.0.x branch

pjfanning:
also updates job for 1.0 doc publishing to use v1.0.2 (latest release)

PR #330, opened 2024-01-27
chore: Make use of statefulMap instead.

He-Pin:
Motivation:
Make use of statefulMap

### DIFF
--- a/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/util/AnnotateLast.scala
+++ b/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/util/AnnotateLast.scala
@@ -16,7 +16,7 @@ package org.apache.pekko.stream.connectors.google.util
 import org.apache.pekko
 import pekko.NotUsed
 import pekko.annotation.InternalApi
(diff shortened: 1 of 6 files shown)


google-common/src/main/scala/org/apache/pekko/stream/connectors/google/util/AnnotateLast.scala
```diff
@@ -47,18 +47,13 @@ private[google] object AnnotateLast {
 
   def apply[T]: Flow[T, MaybeLast[T], NotUsed] =
     Flow[T]
-      .map(Some(_))
-      .concat(Source.single(None))
```
He-Pin, line 51: some kind of `fs`'s `noneTerminate` operator.

pjfanning (approved):
lgtm - but can't be merged util we set up the 1.1 build - basically #325 plus a tag that I will add after 325 is merged

pjfanning:
v1.1 commits are now allowed

He-Pin:
@pjfanning @mdedetrich can this be merged now？

mdedetrich (approved):
lgtm and yes it can since we are on 1.1.0 now

PR #331, opened 2024-01-27
mima compare version

pjfanning:
we are getting issues in `main` branch builds with mima checks - since the v1.1.0-M0 git tag was added

see https://github.com/apache/incubator-pekko-connectors/actions/runs/7678675966/job/20928810106

the issue is because the current setup uses the git tags and expects there to be releases for every git tag

We can test with "1.0.2" for the time being

I raised https://github.com/sbt/sbt-dynver/issues/283 to possibly do a change there so that we could go back to using the `previousStableVersion` sbt variable.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -441,6 +441,8 @@ lazy val `doc-examples` = project
     publish / skip := true,
     Dependencies.`Doc-examples`)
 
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
Lgtm

PR #395, opened 2024-01-27
Enable inliner

mdedetrich:
Enable the inliner since this is now 1.1.x. @pjfanning I know you were sensitive about explicitly setting this to `true` to make it more clear but doing so disables the knob/cli flag at https://github.com/pjfanning/sbt-pekko-build/blob/main/src/main/scala/com/github/pjfanning/pekkobuild/PekkoInlinePlugin.scala#L56 and I don't think explicitly setting `pekkoInlineEnabled` is having much of an effect either way on developers.

If you still want me to explicitly set it to `true` then let me know and ill update the PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,9 +9,6 @@
 
 import net.bzzt.reproduciblebuilds.ReproducibleBuildsPlugin.reproducibleBuildsCheckResolver
 
-// pekkoInlineEnabled should be set to true when we start building 1.1.x builds
-ThisBuild / pekkoInlineEnabled := false
-
 sourceDistName := "apache-pekko-connectors"
 sourceDistIncubating := true
 ThisBuild / resolvers += Resolver.ApacheMavenSnapshotsRepo


pjfanning (approved):
lgtm

mdedetrich:
So although I have tested this in the past already so I am 99% sure there shouldn't be an issue, I will try and make every single CI for each connector pass before merging just to be ultra sure.

mdedetrich:
I will wait until https://github.com/apache/incubator-pekko-connectors/pull/398 is resolved before merging this

He-Pin:
I'd think explicit is better

mdedetrich:
> I'd think explicit is better

It removes the ability to disable the inliner locally which is useful if you want to debug issues

He-Pin:
> > I'd think explicit is better
> 
> It removes the ability to disable the inliner locally which is useful if you want to debug issues

Maybe we can do this

    pekkoInlineEnabled := !sys.props.contains("pekko.no.inline")

mdedetrich:
> Maybe we can do this
> 
> ```
> pekkoInlineEnabled := !sys.props.contains("pekko.no.inline")
> ```

Then we have to repeat this in every single Pekko project which may have inconsistency issues which defeats the whole point of putting it in sbt-pekko-build, i.e. https://github.com/pjfanning/sbt-pekko-build/blob/main/src/main/scala/com/github/pjfanning/pekkobuild/PekkoInlinePlugin.scala#L55-L57

When we are able to release artifacts from CI then we can reverse the defaults so that inline isn't enable by default on local machines

He-Pin (approved):
Lgtm

He-Pin:
> > Maybe we can do this
> > 
> > ```
> > pekkoInlineEnabled := !sys.props.contains("pekko.no.inline")
> > ```
> 
> Then we have to repeat this in every single Pekko project which may have inconsistency issues and also defeats the whole point of putting it in sbt-pekko-build, i.e. https://github.com/pjfanning/sbt-pekko-build/blob/main/src/main/scala/com/github/pjfanning/pekkobuild/PekkoInlinePlugin.scala#L55-L57 so it can be shared in all Pekko projects.
> 
> When we are able to release artifacts from CI then we can reverse the defaults so that inline isn't enable by default on local machines

Let's DRY

mdedetrich:
@He-Pin There is also a related issue on this https://github.com/pjfanning/sbt-pekko-build/issues/9

jxnu-liguobin (approved):

mdedetrich:
I am going to go ahead and merge this, I wanted to try and get a successful run of `mqtt` but its well known that this test is incredibly failure and its hit rate of success is incredibly low. Even though the test is failing there is no observable difference between this PR's `mqtt` tests failing and in general the `mqtt` tests failing.

Issue #409, opened 2024-01-28
Geode test failures in main


mdedetrich:
Probably to do with the latest round of updates from scala-steward but PR's are now getting this

```
An exception or error caused a run to abort: Requires Logback logger for [], it was a [org.slf4j.helpers.NOPLogger] 
java.lang.IllegalArgumentException: Requires Logback logger for [], it was a [org.slf4j.helpers.NOPLogger]
	at org.apache.pekko.stream.connectors.testkit.LogbackUtil$.getLogbackLoggerInternal(LogbackUtil.scala:46)
	at org.apache.pekko.stream.connectors.testkit.LogbackUtil$.getLogbackLogger(LogbackUtil.scala:32)
	at org.apache.pekko.stream.connectors.testkit.CapturingAppender$.get(CapturingAppender.scala:37)
	at org.apache.pekko.stream.connectors.testkit.scaladsl.LogCapturing.$init$(LogCapturing.scala:50)
	at docs.scaladsl.GeodeBaseSpec.<init>(GeodeBaseSpec.scala:31)
	at docs.scaladsl.GeodeFiniteSourceSpec.<init>(GeodeFiniteSourceSpec.scala:25)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.ReflectAccess.newInstance(ReflectAccess.java:128)
	at java.base/jdk.internal.reflect.ReflectionFactory.newInstance(ReflectionFactory.java:347)
	at java.base/java.lang.Class.newInstance(Class.java:645)
	at org.scalatest.tools.Runner$.genSuiteConfig(Runner.scala:1403)
	at org.scalatest.tools.Runner$.$anonfun$doRunRunRunDaDoRunRun$8(Runner.scala:1202)
	at scala.collection.immutable.List.map(List.scala:246)
	at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:1201)
[...]

PR #411, opened 2024-01-28
fix geode test failures

pjfanning:
see #409 

problem started with https://github.com/apache/incubator-pekko-connectors/pull/404/ - I failed to spot the Geode test failures when I merged that

Adding logback dependency fixes the issue


He-Pin (approved):
Lgtm

PR #412, opened 2024-01-28
[DO NOT MERGE] run tests with latest pekko snapshots

pjfanning:


He-Pin:
@inline def asJava: CompletionStage[T] = javaapi.FutureConverters.asJava(f)


We should remove the inline def in pekko.

PR #426, opened 2024-01-29
chore: Do nightly build at 3 AM.

He-Pin:
Motivation:
Do it at 1 AM everyday, then we can check the snapshot just after the pekko-core snapshot published.

### DIFF
--- a/.github/workflows/nightly-builds.yaml
+++ b/.github/workflows/nightly-builds.yaml
@@ -2,7 +2,7 @@ name: Nightly Builds
 
 on:
   schedule:
-    - cron: "0 0 * * *"
+    - cron: "0 3 * * *"
   workflow_dispatch:
 
 permissions: {}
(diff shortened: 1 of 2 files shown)


mdedetrich:
If we are going to stage it this way we should http at 1am, grcp ar 2am and connectors at 3am so all of the dependencies line up. @pjfanning wdyt?

He-Pin:
let't me change to 3 am then.

He-Pin:
Github Action need some kind of action dependencies where actions can depends on each other.

mdedetrich:
I will approve this since this makes sense, @He-Pin Can you make a PR to change the time of pekko-http to 01:00 and pekko-grpc to 02:00

mdedetrich (approved):
lgtm, I will approve this since this makes sense, @He-Pin Can you make a PR to change the time of pekko-http to 01:00 and pekko-grpc to 02:00

laglangyue:
LGTM

PR #445, opened 2024-02-06
influxdb client 2.23

pjfanning:
v2.24 causes issues but this is good enough as a start

https://github.com/apache/incubator-pekko-connectors/pull/380


nvollmar (approved):
lgtm

PR #447, opened 2024-02-06
google-cloud-pubsub 1.122.2

pjfanning:
newer versions cause of protobuf compile issues but this version works and improves out dependency situation


He-Pin (approved):
Lgtm

PR #448, opened 2024-02-06
increase Xmx in builds

pjfanning:
we are getting some build failures but more warnings about GC etc during builds


nvollmar (approved):
lgtm

PR #453, opened 2024-02-07
use latest pravega image

pjfanning:


nvollmar (approved):
lgtm

pjfanning:
I'll merge since the failure is in mqtt-streaming and that is not affected by a pravega image change.

PR #466, opened 2024-02-09
Remove Scala 3 type inference workaround

mdedetrich:
(cherry picked from commit 71141e1294410871e15e579ed05291a0296077c1)


raboof (approved):
I'm not opposed to this change, so LGTM, but I'm not sure it is useful to backport so aggressively.

mdedetrich:
> I'm not opposed to this change, so LGTM, but I'm not sure it is useful to backport so aggressively.

Point taken, for me its more about making sure the branches are clean for the versions of Pekko we still support but I can see how it can be considered excessive.

Issue #480, opened 2024-02-12
[solr] upgrade to solr 9


pjfanning:
We have a compile issue with the Solr 9 libs. The `getIdField` method has been removed.

The Solr team suggested assuming the ID field is "id" because that is the norm anyway.

https://lists.apache.org/thread/fr3o9nntr0f4vj319rnm0zhxhqyyfyy2

We could maintain Solr 8 lib support by using reflection to see if the `getIdField` method is available and call it if found. If the method is not there, then just act as if the method was called and returned "id".

PR #481, opened 2024-02-13
slightly better support for Solr 9 client - (but full support won't be possible until we use Java 11 to build)

pjfanning:
* relates to #480 
* use more robust way to get the router field - getIdField was removed in Solr 9
* turns out Solr 9 needs Java 11
* some of the code has changed in Solr 9 so this code won't work with Solr 9 but it gets closer to being able to support Solr 9
* I might create my own fork to support Solr 9 (requiring a Java 11 build)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -430,8 +430,8 @@ object Dependencies {
         ExclusionRule("software.amazon.awssdk", "netty-nio-client")),
       "org.apache.pekko" %% "pekko-http" % PekkoHttpVersion) ++ Mockito)
 
-  val SolrjVersion = "9.4.1"
-  val SolrVersionForDocs = "9_4"
+  val SolrjVersion = "9.5.0"
+  val SolrVersionForDocs = "9_5"
 
   val Solr = Seq(
     libraryDependencies ++= Seq(
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -430,8 +430,8 @@ object Dependencies {
         ExclusionRule("software.amazon.awssdk", "netty-nio-client")),
       "org.apache.pekko" %% "pekko-http" % PekkoHttpVersion) ++ Mockito)
 
-  val SolrjVersion = "8.11.3"
-  val SolrVersionForDocs = "8_11"
+  val SolrjVersion = "9.4.1"
+  val SolrVersionForDocs = "9_4"
 
   val Solr = Seq(
     libraryDependencies ++= Seq(
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -430,8 +430,8 @@ object Dependencies {
         ExclusionRule("software.amazon.awssdk", "netty-nio-client")),
       "org.apache.pekko" %% "pekko-http" % PekkoHttpVersion) ++ Mockito)
 
-  val SolrjVersion = "9.4.1"
-  val SolrVersionForDocs = "9_4"
+  val SolrjVersion = "8.11.3"
+  val SolrVersionForDocs = "8_11"
 
   val Solr = Seq(
     libraryDependencies ++= Seq(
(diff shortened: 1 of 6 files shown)


project/Dependencies.scala
epugh, line 1: 9.5 was just released ;-)

epugh (commented):
LGTM

project/Dependencies.scala
pjfanning: thanks @epugh -- looks like solrj 9.5 removed one the the new methods that I started to use

csc.getZkStateReader

pjfanning:
Apologies - this seemed to be building ok but now I am finding that some of the test code doesn't compile - so I'll need to investigate that. I will need to get back to this later.

pjfanning:
Sticking with solr client v8 because solr v9 needs Java 11.

epugh:
> Sticking with solr client v8 because solr v9 needs Java 11.

what version of Java does Pekko need?  Solr 10 will probably be Java 17 minimum version, will that be an issue for Pekko?

pjfanning:
> > Sticking with solr client v8 because solr v9 needs Java 11.
> 
> what version of Java does Pekko need? Solr 10 will probably be Java 17 minimum version, will that be an issue for Pekko?

We still support Java 8. I might create my own Solr 9 based Pekko connector in a fork.

epugh:
BTW, are you familiar with https://github.com/inoio/solrs ??

pjfanning:
Is this something that can be merged?

solr/src/main/scala/org/apache/pekko/stream/connectors/solr/impl/SolrFlowStage.scala
```diff
@@ -114,8 +113,12 @@ private final class SolrFlowLogic[T, C](
 
       message.routingFieldValue.foreach { routingFieldValue =>
         val routingField = client match {
-          case csc: CloudSolrClient =>
-            Option(csc.getIdField)
+          case csc: CloudSolrClient => {
+            val docCollection = Option(csc.getZkStateReader.getClusterState.getCollectionOrNull(collection))
```
samueleresca, line 117: minor: I think you can directly access the docCollection using `csc.getZkStateReader.getCollection`. The underlying call is checking for the `clusterState` and calls the `getCollectionOrNull`

pjfanning: thanks - change made

solr/src/main/scala/org/apache/pekko/stream/connectors/solr/impl/SolrFlowStage.scala
```diff
@@ -114,8 +113,12 @@ private final class SolrFlowLogic[T, C](
 
       message.routingFieldValue.foreach { routingFieldValue =>
         val routingField = client match {
-          case csc: CloudSolrClient =>
-            Option(csc.getIdField)
+          case csc: CloudSolrClient => {
+            val docCollection = Option(csc.getZkStateReader.getCollection(collection))
+            docCollection.flatMap { dc =>
+              Option(dc.getRouter.getRouteField(dc))
+            }
```
Roiocam, line 120:
```suggestion
          case csc: CloudSolrClient => {
            val docCollection = Some(csc.getZkStateReader.getCollection(collection))
            docCollection.flatMap { dc =>
              Some(dc.getRouter.getRouteField(dc))
            }
```

`Some` and `None` is a group of subclasses of `Option`, in Scala, we recommend using `Some` rather than directly use `Option`

mdedetrich: Agreed, unless you are expecting values like `csc.getZkStateReader.getCollection(collection)` to be `null`

pjfanning: The whole point is the worry that this Java code could return null. The use of `Option` is deliberate here.

mdedetrich (approved):
lgtm. Will do solr 9 when we split connectors into their own repo so we can bump the major version (since solr 9 requires JDK 11)

Issue #484, opened 2024-02-14
jms consumer leak when exceptions occur


pjfanning:
see https://github.com/akka/alpakka/issues/3038

pjfanning:
https://github.com/apache/incubator-pekko-connectors/pull/485 merged - I'll keep an eye on the tests as they might be brittle

PR #485, opened 2024-02-15
[jms] close sessions when exceptions happen

pjfanning:
relates to #484

### DIFF
--- a/jms/src/main/scala/org/apache/pekko/stream/connectors/jms/impl/JmsConnector.scala
+++ b/jms/src/main/scala/org/apache/pekko/stream/connectors/jms/impl/JmsConnector.scala
@@ -145,7 +145,7 @@ private[jms] trait JmsConnector[S <: JmsSession] {
   }
 
   private def handleRetriableException(ex: Throwable): Unit = {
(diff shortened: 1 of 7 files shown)


nvollmar (approved):
lgtm

mdedetrich:
> relates to #484
> 
> not sure how straightforward this is to unit test - would probably involve opening up a lot of innards of the code

I would be more comfortable if we can somehow test this change as it would also allow us to account for regressions since being a resource leak its quite critical and we want to make sure that it works as intended.

He-Pin (approved):
Lgtm

As it close connection before reconnection

pjfanning:
I'll try to add some test coverage. https://github.com/pjfanning/JmsWrapper is WIP

pjfanning:
@mdedetrich the test change sin JmsProducerRetrySpec fail with the existing code and pass with the new code that closes the JMS sessions in the retriableException code path.

mdedetrich (approved):
lgtm

pjfanning:
I'll keep an eye on the tests because I suspect that the closes could be asynch and may need eventually blocks.

Issue #61, opened 2023-03-16
hbase CI build is disabled


pjfanning:
See https://github.com/apache/incubator-pekko-connectors/blob/main/.github/workflows/check-build-test.yml

Background in https://github.com/akka/alpakka/issues/2185

Ideally, should be fixed

PR #499, opened 2024-02-18
Revert "Update hbase-common, hbase-shaded-client to 1.7.2"

raboof:
Reverts apache/incubator-pekko-connectors#416

* https://github.com/apache/incubator-pekko-connectors/issues/61
* https://github.com/apache/incubator-pekko-connectors/discussions/443

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -16,6 +16,9 @@ updates.pin  = [
 updates.ignore = [
   { groupId = "com.lightbend.paradox", artifactId = "sbt-paradox-project-info" }
   { groupId = "com.lightbend.paradox", artifactId = "sbt-paradox "}
+  # https://github.com/apache/incubator-pekko-connectors/issues/61
+  { groupId = "org.apache.hbase" }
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
Lgtm

pjfanning (commented):
I am not an expert on Scala steward but it seems like the config changes might be wrong. I think it would work better to put these changes in the updates.ignore section since they don't specify versions - I think updates.pin is based on specifying a version or partial version.

mdedetrich (approved):
LGTM, but maybe we should pin it to 1.7.x instead of totally disabling hbase updates?

raboof:
> I am not an expert on Scala steward but it seems like the config changes might be wrong. I think it would work better to put these changes in the updates.ignore section since they don't specify versions - I think updates.pin is based on specifying a version or partial version.

Ah, probably, updated

> LGTM, but maybe we should pin it to 1.7.x instead of totally disabling hbase updates?

If we don't have CI then I'm not sure we should make changes, including minor updates.

PR #505, opened 2024-02-22
Move sbt-paradox to pinned updates

mdedetrich:
We can still use sbt-paradox/sbt-paradox-project-info, just needs to be 0.9.x series


nvollmar (approved):
lgtm

PR #514, opened 2024-02-25
jms build issue

pjfanning:
* build of main branch is broken (CI only) - the build works on my laptop
* I'm not sure if a corrupt version of the jms jar is being used
* none of the recent merges seem to be the cause - I've tried reverting them

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -312,15 +312,12 @@ object Dependencies {
 
   val Jms = Seq(
     libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


mdedetrich:
Can we drop the update of the build.sbt version in the `Update build.properties` commit and let scala-steward do it (or in a separate PR).

Otherwise it seems like its necessary to fix the build which I don't think is the case?

mdedetrich (approved):
lgtm, the jms test passed

PR #522, opened 2024-03-01
fix nightly doc publish

pjfanning:
Copied working setup from pekko-http build.

See https://github.com/apache/incubator-pekko-connectors/actions/runs/8104621838/job/22151574091

Caused by Java 8 unfriendly jars


nvollmar (approved):
lgtm

PR #524, opened 2024-03-01
still trying to fix publish job

pjfanning:


.github/workflows/link-validator.yml
```diff
@@ -22,6 +22,9 @@ jobs:
           distribution: temurin
           java-version: 11
 
+      - name: Install Coursier
```
mdedetrich, line 25: Why do we need this exactly, typically we don't need to install coursier?

pjfanning: got this failure in a recent run on this PR - https://github.com/apache/incubator-pekko-connectors/actions/runs/8113758259/job/22187099743 and I even reran it and got the same result

mdedetrich: Oh its due to link validator, the PR title is misleading then since its talking about publish job (which has nothing to do with link validation)

PR #536, opened 2024-03-03
Update sbt-pekko-paradox

mdedetrich:
New version does not require all of the workarounds

Also ran

```
sbt -Dpekko.genjavadoc.enabled=true "docs/paradox; unidoc"
```

locally to confirm that it works


pjfanning (approved):
lgtm

PR #537, opened 2024-03-03
Fix excludes for sbt-site-paradox

mdedetrich:
The previous [PR](https://github.com/apache/incubator-pekko-connectors/pull/536) forgot to fix the excludes for `sbt-site-paradox`.

Ordinarily we would fix this in https://github.com/apache/incubator-pekko-sbt-paradox/blob/main/build.sbt#L76-L88 however it seems to be only pekko-connectors that uses `sbt-site-paradox` and since its an sbt plugin that [automatically loads](https://github.com/sbt/sbt-site/blob/main/core/src/main/scala/com/typesafe/sbt/site/SitePlugin.scala#L9) adding it to pekko-sbt-paradox isn't appropriate (this is also the reason why we have cases in pekko-connectors where the plugin is [explicitly disabled](https://github.com/apache/incubator-pekko-connectors/blob/main/build.sbt#L22))


pjfanning (approved):
lgtm

pjfanning (approved):
lgtm

PR #539, opened 2024-03-03
[DRAFT] try to test with slf4jv2

pjfanning:


project/Dependencies.scala
```diff
@@ -40,9 +40,11 @@ object Dependencies {
   val hoverflyVersion = "0.14.1"
   val scalaCheckVersion = "1.17.0"
 
-  val LogbackForSlf4j1Version = "1.2.13"
-  val LogbackForSlf4j2Version = "1.3.14"
-  val LogbackVersion = if (PekkoBinaryVersion == "1.0") LogbackForSlf4j1Version else LogbackForSlf4j2Version
+  // Legacy versions support Slf4J v1 for compatibility with older libs
```
Roiocam, line 43: Who is dependent on Slf4J v1? Seems like slf4j2 is backward compatible.  https://www.slf4j.org/faq.html#compatibility

Should we remove logback 1.2?

pjfanning: The tests for solar and kudu fail if we force them to use latest slf4j and logback. People are welcome to try to fix this.

Roiocam: I see, I found those failed tests and commented on them here as an investigation entrance:

- https://github.com/apache/pekko-connectors/actions/runs/8690073541/job/23829155899
- https://github.com/apache/pekko-connectors/actions/runs/8690073541/job/23829160005

pjfanning: it isn't a compile problem - that run is from experimental changes I made. The tests fail due to log capture not working but you can get the tests to compile.

Roiocam: Thanks for the input.

mdedetrich (approved):
lgtm

PR #587, opened 2024-03-27
need build to be able to access snapshot jars

pjfanning:
see https://github.com/apache/pekko-connectors/actions/runs/8446006883/job/23134119754

no longer need to access staging jars


nvollmar (approved):
lgtm

PR #592, opened 2024-04-02
AMQP: Avoid allocating a new ByteString on each consumed message

dimamo5:
Avoids copying a new array and a new allocation by using the `fromArrayUnsafe` method. It should be ok to build from this unsafe method as long as `body` is not exposed to the outside.


pjfanning:
Thanks @dimamo5 but I am not sure that this is necessarily a good idea. How can we be sure that AMQP provider does not reuse byte arrays? We can only use fromArrayUnsafe if we are 100% sure that the byte array won't be reused. Even if 1 AMQP solution can be proved not to reuse arrays, can we be sure that all AMQP solutions don't?

dimamo5:
The critical bit is the implementation of `com.rabbitmq.client.Channel` used in the `DefaultConsumer`. The default `ChannelN` and `RecoveryAwareChannelN` from rabbitmq client do not reuse byte arrays, since they eventually call https://github.com/rabbitmq/rabbitmq-java-client/blob/ca510a078092a3c64b1219feac527bc738b63313/src/main/java/com/rabbitmq/client/impl/CommandAssembler.java#L145 that performs a `System.arraycopy`.

Regarding the AMQP provider, I don't think this connector allows for customization of the `channel` directly from the provider so either one of the abovementioned channels will be used.

pjfanning:
RabbitMQ is not the only AMQP implementation. We could make the ByteString behaviour configurable but I don't think we can hardcore this change.

pjfanning (dismissed):
Blocking merge while this change is debated.

pjfanning:
@dimamo5 I added an `avoidArrayCopy` setting that defaults to false (existing behaviour). I pushed this change to your PR, if you don't mind. Could you have a look?

If this approach is ok, new tests will need to be added - that test with avoidArrayCopy=true.

dimamo5:
@pjfanning sorry for the slow response. This is what I was looking for. Thank you.

pjfanning:
@dimamo5 I've added code so that message writing can also avoid copying the array data for the ByteStrings - also based on `settings.avoidArrayCopy`

I'm pretty busy with other Pekko stuff. Would you have time to add some test cases that say duplicate existing test cases for read and write cases but where `settings.avoidArrayCopy` is set to true?

pjfanning:
I wrote most of the code in this PR so I think I won't be able to approve it.

raboof (approved):
LGTM.

Weird with the MiMa false positives.

We should document when it is safe to enable this option, i.e. when the AMQP implementation guarantees that it does not re-use the buffer passed in here, and perhaps list AMQP implementations that are known to be safe like this?

amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/AmqpConnectorSettings.scala
```diff
@@ -28,6 +28,7 @@ import scala.concurrent.duration._
 sealed trait AmqpConnectorSettings {
   def connectionProvider: AmqpConnectionProvider
   def declarations: immutable.Seq[Declaration]
+  def avoidArrayCopy: Boolean
```
He-Pin, line 31: Maybe just copyArray?

He-Pin (approved):
Lgtm

amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/impl/AmqpRpcFlowStage.scala
```diff
@@ -105,7 +105,14 @@ private[amqp] final class AmqpRpcFlowStage(writeSettings: AmqpWriteSettings, buf
                 body: Array[Byte]): Unit =
               consumerCallback.invoke(
                 new CommittableReadResult {
-                  override val message = ReadResult(ByteString(body), envelope, properties)
+                  override val message = {
+                    val byteString = if (settings.avoidArrayCopy) {
+                      ByteString.fromArrayUnsafe(body)
+                    } else {
+                      ByteString(body)
+                    }
```
laglangyue, line 113:
```scala
         val byteString = if (settings.avoidArrayCopy) 
                         ByteString.fromArrayUnsafe(body)
                     else 
                          ByteString(body)
```

amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/impl/AmqpSourceStage.scala
```diff
@@ -93,10 +93,16 @@ private[amqp] final class AmqpSourceStage(settings: AmqpSourceSettings, bufferSi
               envelope: Envelope,
               properties: BasicProperties,
               body: Array[Byte]): Unit = {
+            val byteString = if (settings.avoidArrayCopy) {
+              ByteString.fromArrayUnsafe(body)
+            } else {
+              ByteString(body)
+            }
```
laglangyue, line 100: same as above

amqp/src/main/mima-filters/1.1.x.backward.excludes/avoidArrayCopy.backwards.excludes
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# added avoidArrayCopy setting
+ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.stream.connectors.amqp.AmqpConnectorSettings.avoidArrayCopy")
```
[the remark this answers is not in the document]
pjfanning: I don't know why mima treats new fields as issues but it fails the build if you don'y add these excludes.

amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/AmqpConnectorSettings.scala
pjfanning: With configs, I prefer the default to be false. If I change this to be copyArray, I'll have to make the default true - because is the current 1.0.0 behaviour.

laglangyue:
LGTM

amqp/src/main/mima-filters/1.1.x.backward.excludes/avoidArrayCopy.backwards.excludes
```diff
@@ -0,0 +1,23 @@
...
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.connectors.amqp.AmqpReplyToSinkSettings.this")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.connectors.amqp.AmqpWriteSettings.this")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.connectors.amqp.NamedQueueSourceSettings.this")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.connectors.amqp.TemporaryQueueSourceSettings.this")
```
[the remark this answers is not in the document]
pjfanning: I suspect that most if not all AMQP client libs should allow you to set this new config to avoid the copy but it's impossible to know how they all behave. That's why I'm being careful and making the change non-default. Our CI tests use RabbitMQ. I wouldn't rush out and endorse this setting for other AMQP services - users would need to carefully test this setting before going into production with it.

amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/AmqpConnectorSettings.scala
JD557: You can also use something like `reuseByteArray` (if you want to use a single verb instead of `avoid[Verb]`)

amqp/src/main/mima-filters/1.1.x.backward.excludes/avoidArrayCopy.backwards.excludes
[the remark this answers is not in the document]
raboof: yes, I was just trying to understand why MiMa was flagging these changes in the first place. In this case I think it has to do with possible problems when traits extend each other, which is not relevant here. Filed https://github.com/lightbend-labs/mima/issues/817 .

amqp/src/main/mima-filters/1.1.x.backward.excludes/avoidArrayCopy.backwards.excludes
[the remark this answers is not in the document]
raboof: (this is already fixed in MiMa but not yet released, https://github.com/lightbend-labs/mima/pull/778)

samueleresca (approved):
Looks good to me

raboof:
dynamodb was failing with `[error] warnings found and -Werror specified`, re-running those

amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/AmqpConnectorSettings.scala
pjfanning: thanks @JD557, I changed to use the name you suggested

pjfanning:
> dynamodb was failing with `[error] warnings found and -Werror specified`, re-running those

This issue was in main branch. It was fixed with #708

Issue #523, opened 2024-03-01
investigate issues with latest AWS libs causing is issues in the SQS connector


pjfanning:
* the lib is upgraded generally but with the SQS connector, we keep the older 2.20.x jars (the most recent jar that works)
* see https://github.com/apache/incubator-pekko-connectors/actions/runs/8058964398/job/22012607914
* PR that upgrades the AWS lib but breaks lots of test

Example
```
[info] - should ask for 'All' attributes set in the settings *** FAILED *** (23 milliseconds)
[info]   java.util.concurrent.ExecutionException: software.amazon.awssdk.services.sqs.model.SqsException: Service returned HTTP status code 400 (Service: Sqs, Status Code: 400, Request ID: null)
[info]   at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
[info]   at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1928)
[info]   at org.apache.pekko.stream.connectors.sqs.scaladsl.DefaultTestContext.randomQueueUrl(DefaultTestContext.scala:62)
[info]   at org.apache.pekko.stream.connectors.sqs.scaladsl.DefaultTestContext.randomQueueUrl$(DefaultTestContext.scala:59)
[info]   at docs.scaladsl.SqsSourceSpec.randomQueueUrl(SqsSourceSpec.scala:47)
[info]   at docs.scaladsl.SqsSourceSpec$IntegrationFixture.$init$(SqsSourceSpec.scala:54)
[info]   at docs.scaladsl.SqsSourceSpec$$anon$5.<init>(SqsSourceSpec.scala:107)
[info]   at docs.scaladsl.SqsSourceSpec.$anonfun$new$5(SqsSourceSpec.scala:107)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.flatspec.AnyFlatSpecLike$$anon$5.apply(AnyFlatSpecLike.scala:1832)
[...]

PR #601, opened 2024-04-08
upgrade sqs lib

pjfanning:
using latest docker image allows us to upgrade the sqs lib
see #523


raboof (approved):
cool, great find!

Roiocam (approved):
LGTM, let us see what happen in CI checks.

PR #607, opened 2024-04-12
update license header

pjfanning:
* this code is not taken from Akka - so it shouldn't have the Akka specific license header
* the original code is Apache licensed


nvollmar (approved):
lgtm

PR #625, opened 2024-04-22
upgrade wiremock and its dependencies due to CVEs

pjfanning:
see https://github.com/apache/pekko-management/pull/208

aim is to reduce the number of security issues listed in https://github.com/apache/pekko-connectors/security/dependabot


samueleresca (approved):
looks good to me.

PR #626, opened 2024-04-23
Using apache kudu docker images and update kudu-client version

laglangyue:
## movtication
we use the kudu images but not from apache，so update it, and update the version
Although I am not an expert in kudu, I am capable of solving this problem. I did some digging and found that the new version of kudu will try TLS connection, so the connection failed

## test
test it locally ok

## future
I add   `encryptionPolicy(EncryptionPolicy.OPTIONAL)` to client temporaryly.
It need to some improvment for client. 
In my opinion, we should provide a way for users to pass us a client that can be customized by the user

```java
  public enum EncryptionPolicy {
    // Optional, it uses encrypted connection if the server supports it
    // but it can connect to insecure servers too.
    OPTIONAL,
    // Only connects to remote servers that support encryption, fails
    // otherwise. It can connect to insecure servers only locally.
    REQUIRED_REMOTE,
    // Only connects to any server, including on the loopback interface,
    // that support encryption, fails otherwise.
    REQUIRED,
  }
```

### DIFF
--- a/kudu/src/main/scala/org/apache/pekko/stream/connectors/kudu/KuduClientExt.scala
+++ b/kudu/src/main/scala/org/apache/pekko/stream/connectors/kudu/KuduClientExt.scala
@@ -13,8 +13,9 @@
 
 package org.apache.pekko.stream.connectors.kudu
 
+import org.apache.kudu.client.AsyncKuduClient.EncryptionPolicy
 import org.apache.pekko
-import pekko.actor.{ ClassicActorSystemProvider, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider }
+import pekko.actor.{ClassicActorSystemProvider, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider}
 import org.apache.kudu.client.KuduClient
 
 /**
@@ -23,9 +24,10 @@ import org.apache.kudu.client.KuduClient
 final class KuduClientExt private (sys: ExtendedActorSystem) extends Extension {
   val client = {
     val masterAddress = sys.settings.config.getString("pekko.connectors.kudu.master-address")
-    new KuduClient.KuduClientBuilder(masterAddress).build
+    new KuduClient.KuduClientBuilder(masterAddress)
+      .encryptionPolicy(EncryptionPolicy.OPTIONAL)
+      .build()
   }
-
   sys.registerOnTermination(client.shutdown())
 }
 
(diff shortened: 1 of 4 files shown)


laglangyue:
@pjfanning 
I know you've made some efforts for Kudu before, what do you think?

laglangyue:
reference to
https://github.com/apache/kudu/blob/master/docker/quickstart.yml

laglangyue:
I check the log about container. 
image

Then I found a issue about kudu TLS
https://issues.apache.org/jira/browse/KUDU-2647

kudu/src/main/scala/org/apache/pekko/stream/connectors/kudu/KuduClientExt.scala
```diff
@@ -13,8 +13,9 @@
 
 package org.apache.pekko.stream.connectors.kudu
 
+import org.apache.kudu.client.AsyncKuduClient.EncryptionPolicy
 import org.apache.pekko
-import pekko.actor.{ ClassicActorSystemProvider, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider }
+import pekko.actor.{ClassicActorSystemProvider, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider}
```
pjfanning, line 18: please fix the scalafmt

kudu/src/main/scala/org/apache/pekko/stream/connectors/kudu/KuduClientExt.scala
```diff
@@ -13,8 +13,9 @@
 
 package org.apache.pekko.stream.connectors.kudu
 
+import org.apache.kudu.client.AsyncKuduClient.EncryptionPolicy
```
pjfanning, line 16: please keep the kudu imports together

project/Dependencies.scala
```diff
@@ -420,7 +418,7 @@ object Dependencies {
       "com.google.jimfs" % "jimfs" % "1.3.0" % Test,
       "org.scalacheck" %% "scalacheck" % scalaCheckVersion % Test,
       "org.scalatestplus" %% scalaTestScalaCheckArtifact % scalaTestScalaCheckVersion % Test) ++
-    wireMockDependencies)
+      wireMockDependencies)
```
pjfanning, line 421: please remove this format change - it will probably cause a scalafmt failure

laglangyue:
I check it again, I add a config for EncryptionPolicy. PTAL @pjfanning 
In fact, implementation of  kudu connector is too simple, and If  I have time, I will submit a pr in futrue.

pjfanning:
Thanks - I've restarted the CI runs

pjfanning (approved):
lgtm

pjfanning:
Thanks @laglangyue - merged. I have labelled this as needing a mention in the release notes due to the new config.

Issue #627, opened 2024-04-24
Investigating the Flaky Test failed fot FTP


laglangyue:
https://github.com/apache/pekko-connectors/actions/runs/8810670241/job/24193635981?pr=624

raboof:
hmm, perhaps one of our test certificates expired? `NotAfter: Wed Apr 24 06:36:56 UTC 2024`

pjfanning:
Seems like an expired cert. The tests seem to fail consistently now. They are not flaky.

pjfanning:
I put the details of the fix for the cert in https://github.com/apache/pekko-connectors/pull/631

PR #629, opened 2024-04-25
gen new cert for ftp test

pjfanning:
see #627


pjfanning:
this change doesn't help and causes new issues - the issue may be in ftp/src/test/resources/ftpd/pure-ftpd.pem as opposed to these files - I'll have a look at that file later

PR #633, opened 2024-04-26
add pekko-connectors-bom

pjfanning:
We have BOMs for Pekko(Core) and Pekko-HTTP.

* https://repo1.maven.org/maven2/org/apache/pekko/pekko-bom_2.13/1.0.2/pekko-bom_2.13-1.0.2.pom
* https://repo1.maven.org/maven2/org/apache/pekko/pekko-http-bom_2.13/1.0.1/pekko-http-bom_2.13-1.0.1.pom

I think it useful to have one for Pekko-Connectors.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,58 +15,60 @@ ThisBuild / resolvers += Resolver.ApacheMavenSnapshotsRepo
 
 ThisBuild / reproducibleBuildsCheckResolver := Resolver.ApacheMavenStagingRepo
 
+lazy val userProjects: Seq[ProjectReference] = List[ProjectReference](
+  amqp,
+  avroparquet,
+  awslambda,
+  azureStorageQueue,
+  cassandra,
+  couchbase,
+  csv,
+  dynamodb,
+  elasticsearch,
+  eventbridge,
+  files,
+  ftp,
+  geode,
+  googleCommon,
+  googleCloudBigQuery,
+  googleCloudBigQueryStorage,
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -434,6 +436,16 @@ lazy val `doc-examples` = project
     publish / skip := true,
     Dependencies.`Doc-examples`)
 
+lazy val billOfMaterials = Project("bill-of-materials", file("bill-of-materials"))
+  .enablePlugins(BillOfMaterialsPlugin)
+  .disablePlugins(MimaPlugin, SitePlugin)
+  .settings(
+    name := "pekko-connectors-bom",
+    licenses := List(License.Apache2),
+    libraryDependencies := Seq.empty,
```
Roiocam, line 445: `libraryDependencies` default value is Seq.empty? if true we could ignore this.

Roiocam (approved):
Looks good, Thanks. 

If we have a BOM that records all modules(more tha this), it will be better, but it will come with higher maintenance costs.

build.sbt
pjfanning: this is needed to fight against the commonSettings in the build - that hardcodes a dependency on pekko-stream.
BOMs don't have dependencies.

Roiocam:
>  My project is easier to maintain if the main pekko repos have BOMs that I can more easily aggregate.

It sounds reasonable.  Let's move forward.

PR #648, opened 2024-05-08
Move AWS SPI code into Pekko repo

pjfanning:
* See https://github.com/apache/pekko-connectors/issues/612
* follow up in https://github.com/apache/pekko-connectors/issues/655


pjfanning:
@mdedetrich I'm thinking of just removing the integration tests. They require real AWS access. There are unit tests that use testcontainers. I think those are enough. Is this ok?

mdedetrich:
> @mdedetrich I'm thinking of just removing the integration tests. They require real AWS access. There are unit tests that use testcontainers. I think those are enough. Is this ok?

So we do actually have a real AWS account that is provided by Apache INFRA and it is currently being used by [S3 tests](https://github.com/apache/pekko-connectors/blob/8c5d65051cfeb3df4a0eb87039c33131b6b21ccc/s3/src/test/scala/org/apache/pekko/stream/connectors/s3/scaladsl/S3IntegrationSpec.scala#L1267-L1282) and there is a [github action that triggers it daily off the main branch](https://github.com/apache/pekko-connectors/blob/e77bc690d85e91134c5a9571d733ea2a66c3de67/.github/workflows/nightly-builds.yaml)

Initially I would suggest trying to enable the tests and if its too much effort we can disable them, file an issue and try and re-enable them later. The reason I am suggesting this is that the core reason for me pushing for the AWS S3 integration tests is we did have an actual case of a regression that wasn't caught since initially there wasn't tests running against it.

mdedetrich:
Oh and I forgot to mention that if you do end up enabling integration tests, you have to work with Apache INFRA so that they can add in the necessary roles/permissions in the AWS account for whatever is being tested.

This does require some back and forth so given that it might make sense to do it in a different PR, but regardless I think we should aim to have these tests in place

mdedetrich:
@pjfanning Is this ready

pjfanning:
I'll leave the integration tests to another time but I have a couple of more things to do with this PR - it is pretty close now.

aws-spi-pekko-http/src/it/scala/org/apache/pekko/stream/connectors/awsspi/dynamodb/ITTestDynamoDB.scala
```diff
@@ -0,0 +1,91 @@
...
+import pekko.stream.connectors.awsspi.{ PekkoHttpAsyncHttpService, TestBase }
+import pekko.util.FutureConverters
+import org.scalatest.concurrent.{ Eventually, Futures, IntegrationPatience }
+import org.scalatest.concurrent.ScalaFutures._
+import org.scalatest.wordspec.AnyWordSpec
+import org.scalatest.matchers.should.Matchers
+import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
+import software.amazon.awssdk.services.dynamodb.model._
+
+class ITTestDynamoDB
```
laglangyue, line 30: It seems that `IT` and `Test` are usually placed at the end

He-Pin:
I will take a look this weekend

build.sbt
```diff
@@ -128,7 +129,13 @@ lazy val amqp = pekkoConnectorProject("amqp", "amqp", Dependencies.Amqp)
 lazy val avroparquet =
   pekkoConnectorProject("avroparquet", "avroparquet", Dependencies.AvroParquet)
 
-lazy val awslambda = pekkoConnectorProject("awslambda", "aws.lambda", Dependencies.AwsLambda)
+lazy val awsSpiPekkoHttp =
+  pekkoConnectorProject("aws-spi-pekko-http", "aws.api.pekko.http", Dependencies.AwsSpiPekkoHttp)
+    .configs(IntegrationTest)
+    .settings(Defaults.itSettings)
+
+lazy val awslambda = pekkoConnectorProject("awslambda", "aws.api.pekko.http", Dependencies.AwsLambda)
```
samueleresca, line 137: Is the moduleName correct? Shoudln't be:

`pekkoConnectorProject("awslambda", "aws.lambda", Dependencies.AwsLambda)`

pjfanning: thanks - I've fixed the module name now

samueleresca (commented):
looks good to me, Can someone else add their review?

pjfanning:
Would anyone have time to review this PR? I'm hoping to finalise the 1.1.0-M1 RC1 over the next week or so.

aws-spi-pekko-http/src/it/scala/org/apache/pekko/stream/connectors/awsspi/dynamodb/ITTestDynamoDB.scala
nvollmar: Agree, should be a post fix

nvollmar (approved):
lgtm
Just test naming should be fixed at some point

samueleresca (approved):
already reviewed. I'm proceeding by approving.

mdedetrich (approved):
lgtm, sorry for taking a while

PR #658, opened 2024-05-16
s3: avoid array copies when dealing with ByteStrings

pjfanning:
the 'unsafe' methods are ok if you know the data is not being mutated

### DIFF
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/auth/package.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/auth/package.scala
@@ -36,7 +36,16 @@ package object auth {
     new String(out)
   }
 
-  @InternalApi private[impl] def encodeHex(bytes: ByteString): String = encodeHex(bytes.toArray)
+  @InternalApi private[impl] def encodeHex(bytes: ByteString): String = {
+    val length = bytes.length
+    val out = new Array[Char](length * 2)
+    for (i <- 0 until length) {
+      val b = bytes(i)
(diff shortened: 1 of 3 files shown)


pjfanning:
@mdedetrich You are the most active contributor to the S3 code base. Do these changes look ok to you?

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/auth/package.scala
```diff
@@ -36,7 +36,16 @@ package object auth {
     new String(out)
   }
 
-  @InternalApi private[impl] def encodeHex(bytes: ByteString): String = encodeHex(bytes.toArray)
+  @InternalApi private[impl] def encodeHex(bytes: ByteString): String = {
+    val length = bytes.length
```
pjfanning, line 40: ByteString implements `IndexedSeq[Byte]` so we can iterate it without converting it to an array.

pjfanning: looking at the ByteStrings code (complex ByteString made up of underlying ByteStrings) the apply(Int) function and the iterator function are not performant.

https://github.com/apache/pekko/blob/c4805f1839ead0cfe31700d6c41928df9fb1056d/actor/src/main/scala-3/org/apache/pekko/util/ByteString.scala#L554-L570

So I'm now thinking of changing this to
```
def encodeHex(bytes: ByteString): String = encodeHex(bytes.toArrayUnsafe)
```

nvollmar (approved):
lgtm

PR #676, opened 2024-05-31
fix link-validator.yml

pjfanning:
always fails - https://github.com/apache/pekko-connectors/actions/workflows/link-validator.yml


Roiocam (approved):
LGTM, seem like the pekko repo do the same thing too.

Issue #677, opened 2024-06-01
fix link validation job


pjfanning:
https://github.com/apache/pekko-connectors/actions/runs/9331434430/job/25686093801

The job had been broken altogether. It is now running but there are a lot of issues.

pjfanning:
```
## HTTP failure response
`[http://fastutil.di.unimi.it/`](http://fastutil.di.unimi.it/%60) status -1 akka.stream.StreamTcpException: Tcp command [Connect(fastutil.di.unimi.it:80,None,List(),Some(10 seconds),true)] failed because of akka.io.TcpOutgoingConnection$$anon$2: Connect timeout of Some(10 seconds) expired
 - ../../../../docs/target/site/docs/pekko-connectors/snapshot/license-report.html

`[http://stephenc.github.com/findbugs-annotations`](http://stephenc.github.com/findbugs-annotations%60) status 404 Not Found
 - ../../../../docs/target/site/docs/pekko-connectors/snapshot/license-report.html

`[http://stephenc.github.com/jcip-annotations`](http://stephenc.github.com/jcip-annotations%60) status 404 Not Found
 - ../../../../docs/target/site/docs/pekko-connectors/snapshot/license-report.html

`[https://docs.datastax.com/en/developer/java-driver/`](https://docs.datastax.com/en/developer/java-driver/%60) status 404 Not Found
 - ../../../../docs/target/site/api/pekko-connectors/snapshot/org/apache/pekko/stream/connectors/cassandra/javadsl/CassandraSession.html
 - ../../../../docs/target/site/api/pekko-connectors/snapshot/org/apache/pekko/stream/connectors/cassandra/scaladsl/CassandraSession.html

`[https://docs.datastax.com/en/developer/java-driver/4.17/`](https://docs.datastax.com/en/developer/java-driver/4.17/%60) status 404 Not Found
 - ../../../../docs/target/site/docs/pekko-connectors/snapshot/cassandra.html

`[https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/configuration/`](https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/configuration/%60) status 404 Not Found
 - ../../../../docs/target/site/docs/pekko-connectors/snapshot/cassandra.html

`[https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/configuration/reference/`](https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/configuration/reference/%60) status 404 Not Found
[...]

PR #678, opened 2024-06-01
fix some doc issues

pjfanning:
doesn't fix everything in #677 but is a start
* use Java 8 to build like we do when we publish the docs - the Java API links work when Java 8 is used but the incorrect paths are used when Java 11 is used to build the docs
* we have very little control over all the mistakes in the generated license report


nvollmar (approved):
lgtm

PR #683, opened 2024-06-02
reduce size of LegacyFtpsClient

pjfanning:
changes in commons-net 3.11.0 let us get rid of most of the code that we needed to copy in from commons-net


nvollmar (approved):
lgtm

laglangyue (approved):
lgtm

PR #710, opened 2024-06-30
test with pekko 1.0.x

pjfanning:
* main builds already test with pekko 1.1 so repurpose the dedicated pekko1.1 nightly run to now use pekko 1.0 
* also test compile with Java 21

### DIFF
--- a/.github/workflows/nightly-builds.yaml
+++ b/.github/workflows/nightly-builds.yaml
@@ -8,6 +8,35 @@ on:
 permissions: {}
 
 jobs:
+  test-compile:
+    name: Compile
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly-builds.yaml
```diff
@@ -8,6 +8,35 @@ on:
 permissions: {}
 
 jobs:
+  test-compile:
+    name: Compile
+    runs-on: ubuntu-20.04
```
nvollmar, line 13: quite old, maybe we should got through and update the images used in all workflows

nvollmar (approved):
lgtm

Issue #296, opened 2023-12-23
DirectoryChangesSource.java does not compile with Java 21


pjfanning:
The CI build applies -Werror when CI build happens (see project/Common.scala)

DirectoryChangesSource.java uses a com.sun class that has been deprecated.

I tried adding `@SuppressWarnings("deprecation")` in various places but the Werror check still fails.

raboof:
The deprecated flag is:

```
      private final WatchKey watchKey =
          directoryPath.register(
              service,
              new WatchEvent.Kind<?>[] {ENTRY_CREATE, ENTRY_MODIFY, ENTRY_DELETE, OVERFLOW},
              // this is com.sun internal, but the service is useless on OSX without it
              SensitivityWatchEventModifier.HIGH);
```

https://bugs.openjdk.org/browse/JDK-8307097 says:

> The polling based `WatchService` has been changed to ignore these modifiers when registering files to be watched.

It seems there's no alternative...

raboof:
Should we keep this open to decide if/when/how/... we can drop the reference to `SensitivityWatchEventModifier.HIGH`

pjfanning:
Can we keep this open to remind us about the issue? At some stage, Java might announce the class is being removed and we will be forced to react.

raboof:
> Can we keep this open to remind us about the issue? At some stage, Java might announce the class is being removed and we will be forced to react.

Sounds reasonable to me! It probably already affects people who use the module system right now.

PR #713, opened 2024-07-01
disable -Werror check on Java 21

pjfanning:
see #296


raboof (approved):
A shame, I'm not opposed to the change but perhaps we can reduce the scope to just the `file` component like #714?

pjfanning:
close in favour of #714

PR #674, opened 2024-05-27
Adding Jakarta Messaging 3.0 support

samueleresca:
See: #483 

Few notes regarding the PR after spending some time on this:

- Jakarta JMS 3.1 requires Java 11+, see [Jakarta Messaging 3.1.0 Release Review](https://projects.eclipse.org/projects/ee4j.messaging/reviews/3.1.0-release-review). This PR is upgrading to [Jakarta JMS 3.0](https://projects.eclipse.org/projects/ee4j.jms/reviews/3.0.0-release-review) which is the latest release supporting Java 8.
- ActiveMQ Classic is used in the jms connector to perform the integration tests. ActiveMQ classic cannot be upgraded to a version later than `5.16.7` because the later versions don't support Java 8. 
- ActiveMQ Classic 5.16.7 only [supports Javax JMS 1.1](https://activemq.apache.org/components/classic/download/).
- I replaced ActiveMQ Classic with ActiveMQ Artemis, which [supports both the new versions of JMS and also Java 8](https://activemq.apache.org/components/artemis/download/past_releases). I wasn't able to port the following 2 tests using ActiveMQ Artemis:
   - `JmsProducer should retry sending on network failures`
   - `The JMS Transactional Connectors ensure no message loss or starvation when exceptions occur in a stream missing commits`

PR #718, opened 2024-07-03
add jakartams to sbt project list

pjfanning:
Added in #674 but this sbt issue seems to stop the jar being published

see https://repository.apache.org/content/groups/snapshots/org/apache/pekko/ - no jakartams there

* also tidies up imports
* fixes doc build issues - the class names in the new connector clash with those in the jms connector 
* fix mima issue - there is no previous jar to compare against

### DIFF
--- a/jakartams/src/main/scala/org/apache/pekko/stream/connectors/jakartams/impl/JmsInternalMatValues.scala
+++ b/jakartams/src/main/scala/org/apache/pekko/stream/connectors/jakartams/impl/JmsInternalMatValues.scala
@@ -14,10 +14,10 @@
 package org.apache.pekko.stream.connectors.jakartams.impl
 
 import org.apache.pekko
-import org.apache.pekko.NotUsed
-import org.apache.pekko.annotation.InternalApi
-import org.apache.pekko.stream.KillSwitch
-import org.apache.pekko.stream.scaladsl.Source
+import pekko.NotUsed
+import pekko.annotation.InternalApi
+import pekko.stream.KillSwitch
+import pekko.stream.scaladsl.Source
 
 /**
  * Internal API.
--- a/build.sbt
+++ b/build.sbt
@@ -475,7 +475,8 @@ lazy val billOfMaterials = Project("bill-of-materials", file("bill-of-materials"
     bomIncludeProjects := userProjects,
     description := s"${description.value} (depending on Scala ${CrossVersion.binaryScalaVersion(scalaVersion.value)})")
 
-val mimaCompareVersion = "1.0.2"
+private val mimaCompareVersion = "1.0.2"
+private val noMimaChecks = Set("slick", "couchbase3", "jakartams", "aws.api.pekko.http")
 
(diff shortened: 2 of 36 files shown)


pjfanning:
@samueleresca the new #674 classes are breaking the doc build. See the checks for this PR.

It looks hard to fix. Paradox has built-in support for having 2 classes of the same name but one is in javadsl and one is in scaladsl. This support can't handle having a 2nd set of classes with the same name.

Would it be possible to rename all the `Jms` prefixed classes in the new connector to have something like `JakartaMs` or `JakartaMessaging` or `JakartaJms` prefix?

pjfanning:
@samueleresca I think I have found a way to fix the doc issue - just testing it locally right now

jakartams/src/main/scala/org/apache/pekko/stream/connectors/jakartams/impl/JmsInternalMatValues.scala
```diff
@@ -14,10 +14,10 @@
...
 
 import org.apache.pekko
-import org.apache.pekko.NotUsed
-import org.apache.pekko.annotation.InternalApi
-import org.apache.pekko.stream.KillSwitch
-import org.apache.pekko.stream.scaladsl.Source
+import pekko.NotUsed
+import pekko.annotation.InternalApi
+import pekko.stream.KillSwitch
+import pekko.stream.scaladsl.Source
```
He-Pin, line 20: Maybe better if sort the imports

He-Pin (approved):
Lgtm

jakartams/src/main/scala/org/apache/pekko/stream/connectors/jakartams/impl/JmsInternalMatValues.scala
pjfanning: these look sorted to me already

samueleresca:
> @samueleresca the new #674 classes are breaking the doc build. See the checks for this PR.
> 
> It looks hard to fix. Paradox has built-in support for having 2 classes of the same name but one is in javadsl and one is in scaladsl. This support can't handle having a 2nd set of classes with the same name. The issue is if we try to use the FQCN, we will need to include the full package name and this negates the Paradox support for linking the javadsl/scaladsl version of the class depending on whether the UI user has chosen Java or Scala.
> 
> Would it be possible to rename all the `Jms` prefixed classes in the new connector to have something like `JakartaMs` or `JakartaMessaging` or `JakartaJms` prefix?

Why the build wasn't catching this at PR level? (I haven't seen anything failing). Let me come back on this in the next few days. I can rename the classes and give it a try again. Sorry for the inconvenience

pjfanning:
@samueleresca I think this PR fixes the issues. The main build issue was the fact this new module wasn't added to the project list in build.sbt. This means that non-scoped sbt comands will ignore the module. So `sbt jakartams/test` works but `sbt publishM2` ignored jakartams.

Issue #468, opened 2024-02-09
docs.javadsl.MqttFlowTest: bidirectional connection may leak actors


raboof:
It seems to be specifically `establishClientBidirectionalConnectionAndSubscribeToATopic` and `establishServerBidirectionalConnectionAndSubscribeToATopic`

raboof:
(let's close this for now in case #736 was it, and re-open when it turns out it wasn't)

raboof:
https://github.com/apache/pekko-connectors/actions/runs/9892582296/job/27325614127

raboof:
The problem is that, after the test finishes, there are still Streams components 'in flight', indicating some kind of leak in the stream teardown logic.

The component that checks for these remaining running components helpfully products 'dot' graphs to diagnose what is waiting for what. Unfortunately these confuse me even more: it *seems* even 'headSink' is still waiting for input, even though that component should complete after producing its one and only value (and the test completing suggests that this has already happened). That could use additional research: is the component indeed still running, or is the dot output just confusing?

However, while looking at this test again, I noticed that it is still using a topic that is also used in another test. While that doesn't really explain why the components still seem to be running after the test has completed, no good can come of it, so let's make sure those tests are independent with #981.

Of course multiple tests sharing the same topic should still not lead to stream components getting left behind - but if making the tests more independent fixes the symptom of these flaky tests that's already a win in my book.

PR #456, opened 2024-02-08
mqtt-streaming: fix flaky 'duplicate publish given two client connections' test

raboof:
It appears the flakiness was caused by the two connections using the same connection id for different connections in the same session. I'm not intimately familiar with this codebase, but AFAICS it seems this connection id is an internal concern to the component, and not an MQTT concept. I think we should probably provide an API that does not bother the user with this concept at all, and manages non-overlapping connection ids entirely internal to the component - filed #457. I don't get the impression that it was the purpose of the test to test the behaviour in case of overlapping connection ids.

Fixes #454

This PR still fails due to #458 but #460 suggests it should be OK when combined with #459.

(note the diff is much simpler than it looks, use the 'Hide whitespace' under the cog in the diff view to ignore the indent change)

PR #732, opened 2024-07-07
fix: don't use overlapping session ids in MqttFlowTest

raboof:
While looking into #468, I noticed the two failing tests were sharing the same session id, which reminded of #456.

While in this case the two tests aren't sharing the same session, and I haven't investigated the details of this codebase further, I'm curious to see if the problem remains when we use unique session ids.


pjfanning (approved):
lgtm

raboof:
> While in this case the two tests aren't sharing the same session, and I haven't investigated the details of this codebase further, I'm curious to see if the problem remains when we use unique session ids.

Unfortunately, it turns out this was not the problem (https://github.com/apache/pekko-connectors/actions/runs/9827053364/job/27129293765)

Issue #462, opened 2024-02-09
UnixDomainSocketSpec: A Unix Domain Socket should not be able to connect to a non-existent file is flaky


raboof:
```
[info] UnixDomainSocketSpec:
[info] A Unix Domain Socket
[info] - should receive some bytes (34 milliseconds)
[info] - should send and receive more ten times the size of a buffer !!! IGNORED !!!
[info] - should allow the client to close the connection (18 milliseconds)
[info] - should close the server once the client is also closed (1 second, 30 milliseconds)
[info] - should be able to materialize outgoing connection flow more than once (9 milliseconds)
[info] - should not be able to bind to socket in a non-existent directory (7 milliseconds)
[info] - should not be able to connect to a non-existent file *** FAILED *** (31 milliseconds)
[info]   Unexpected successful completion with value [None] (UnixDomainSocketSpec.scala:195)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)
[info]   at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)
[info]   at docs.scaladsl.UnixDomainSocketSpec.newAssertionFailedException(UnixDomainSocketSpec.scala:38)
[info]   at org.scalatest.Assertions.fail(Assertions.scala:933)
[info]   at org.scalatest.Assertions.fail$(Assertions.scala:929)
[info]   at docs.scaladsl.UnixDomainSocketSpec.fail(UnixDomainSocketSpec.scala:38)
[info]   at docs.scaladsl.UnixDomainSocketSpec.$anonfun$new$20(UnixDomainSocketSpec.scala:195)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.wordspec.AnyWordSpecLike$$anon$3.apply(AnyWordSpecLike.scala:1240)
[info]   at org.apache.pekko.stream.connectors.testkit.scaladsl.LogCapturing.withFixture(LogCapturing.scala:70)
[...]

PR #734, opened 2024-07-07
chore: add diagnostics to diagnose UnixDomainSocketSpec

raboof:
Looking at `UnixDomainSocketImpl#outgoingConnection`, the test failure at #462 can/should only happen when both `connectionFlow` and `connectionFinished` complete successfully.

There seems to be potential race between `UnixDomainSocketImpl#connectKey` and `channel.connect` both completing `connectionFinished`, though this seems strange. This PR is meant to add diagnostic logging so next time the test fails we can confirm whether this is indeed what's going on.

### DIFF
--- a/unix-domain-socket/src/main/scala/org/apache/pekko/stream/connectors/unixdomainsocket/impl/UnixDomainSocketImpl.scala
+++ b/unix-domain-socket/src/main/scala/org/apache/pekko/stream/connectors/unixdomainsocket/impl/UnixDomainSocketImpl.scala
@@ -271,20 +271,22 @@ private[unixdomainsocket] object UnixDomainSocketImpl {
     }
   }
 
-  private def connectKey(remoteAddress: JnrUnixSocketAddress,
-      connectionFinished: Promise[Done],
+  private def connectKey(connectionFinished: Promise[Done],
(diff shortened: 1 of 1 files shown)


unix-domain-socket/src/main/scala/org/apache/pekko/stream/connectors/unixdomainsocket/impl/UnixDomainSocketImpl.scala
```diff
@@ -271,20 +271,22 @@ private[unixdomainsocket] object UnixDomainSocketImpl {
...
     cancellable.foreach(_.cancel())
     try {
       connectingChannel.register(sel, SelectionKey.OP_READ, sendReceiveContext)
       val finishExpected = connectingChannel.finishConnect()
+      log.debug(s"Unix domain socket connection succeeded, finish $finishExpected")
       require(finishExpected, "Internal error - our call to connection finish wasn't expected.")
       connectionFinished.trySuccess(Done)
     } catch {
       case NonFatal(e) =>
+        log.debug(s"Unix domain socket connection failed")
```
pjfanning, line 289: no interpolation needed here

pjfanning (approved):
lgtm

PR #735, opened 2024-07-07
chore: avoid scala-steward opening multiple PRs for Jackson

raboof:
As those versions are managed by a single variable in our build


pjfanning (approved):
lgtm

PR #740, opened 2024-07-11
Create publish-1.1-docs.yml

pjfanning:
for publishing docs for v1.1.x


He-Pin (approved):
lgtm

pjfanning:
> LGTM. This would be for a one-off manually triggered publish action, right? I wonder if it'd make sense to use https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs for a reusable workflow.

I'm not against trying to remove duplication. I can look at this again another time.

PR #760, opened 2024-07-31
SQS - Add delaySeconds to SQS batch publishing

### LINKED ISSUE #759: SqsPublishFlow - parameter delaySeconds is not populated in SendMessageBatchRequestEntry
Hi, in library `peko-connectors-sqs-1.x.x`, `delaySeconds` is a property which can be specified on each individual message request (`SendMessageRequest`). However, when these requests are mapped to `SendMessageBatchRequestEntry`, property `delaySeconds` is left out (check [SqsPublishFlow code](https://github.com/apache/pekko-connectors/blob/main/sqs/src/main/scala/org/apache/pekko/stream/connectors/sqs/scaladsl/SqsPublishFlow.scala#L89-L96)).

Is there a workaround for this, or do you think I should raise a pull request with the missing mapping?

Code snippets:

message request
```        
SendMessageRequest
    .builder()
    .queueUrl(queueUrl)
    .delaySeconds(delaySeconds)
    .messageBody(rawJson)
    .messageAttributes(attributeMap.asJava)
    .build()
```
batch request entry mapping
```
    Flow[Iterable[SendMessageRequest]]
      .map { requests =>
        val entries = requests.zipWithIndex.map {
          case (r, i) =>
            SendMessageBatchRequestEntry
              .builder()
              .id(i.toString)
              .messageBody(r.messageBody())
              .messageAttributes(r.messageAttributes())
              .messageGroupId(r.messageGroupId())
              .messageDeduplicationId(r.messageDeduplicationId())
          // NOTE: missing mapping for delaySeconds
              .build()
        }
```

Impact:
It is not possible to specify individual delay on an sqs message, while AWS supports this functionality (check example 2 on the [documentation page](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html#API_SendMessageBatch_Examples)).

llazanja:
Add missing `delaySeconds` property in the mapping from `SendMessageRequest` to `SendMessageBatchRequestEntry`
 
 Fixes #759

### DIFF
--- a/sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
+++ b/sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
@@ -181,9 +181,9 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
   it should "publish batch of SendMessageRequests without individual delays and pull them" taggedAs Integration in {
     new IntegrationFixture {
       // #batch-send-request
-      val messages = for (i <- 0 until 10)
-        yield
-          SendMessageRequest
+      val messages =
+        for (i <- 0 until 10)
+          yield SendMessageRequest
             .builder()
             .messageBody(s"Message - $i")
             .delaySeconds(0)
@@ -203,9 +203,9 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
   it should "publish batch of SendMessageRequests with individual delays and pull them" taggedAs Integration in {
     new IntegrationFixture {
       val messageDelaySeconds = 5
-      val messages = for (i <- 0 until 10)
-        yield
-          SendMessageRequest
+      val messages =
+        for (i <- 0 until 10)
+          yield SendMessageRequest
             .builder()
             .messageBody(s"Message - $i")
             .delaySeconds(messageDelaySeconds) // SQS queue should process messages after 5 seconds
--- a/sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
+++ b/sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
@@ -33,6 +33,7 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
 
   abstract class IntegrationFixture(fifo: Boolean = false) {
     val queueUrl: String = if (fifo) randomFifoQueueUrl() else randomQueueUrl()
+    val messageReceiveMinimumTimeout: Int = 2
     implicit val awsSqsClient: SqsAsyncClient = sqsClient
 
     def receiveMessage(): Message =
@@ -43,18 +44,20 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
         .asScala
         .head
 
-    def receiveMessages(maxNumberOfMessages: Int): Seq[Message] = {
+    def receiveMessages(maxNumberOfMessages: Int, waitTimeSeconds: Int = 0): Seq[Message] = {
       // see https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/model/ReceiveMessageRequest.html
       require(maxNumberOfMessages > 0 && maxNumberOfMessages <= 10, "maxNumberOfMessages must be in 1 to 10")
 
+      val timeoutSeconds = messageReceiveMinimumTimeout + waitTimeSeconds
       val request =
         ReceiveMessageRequest
           .builder()
           .queueUrl(queueUrl)
           .maxNumberOfMessages(maxNumberOfMessages)
+          .waitTimeSeconds(waitTimeSeconds)
           .build()
 
-      awsSqsClient.receiveMessage(request).get(2, TimeUnit.SECONDS).messages().asScala.toSeq
+      awsSqsClient.receiveMessage(request).get(timeoutSeconds, TimeUnit.SECONDS).messages().asScala.toSeq
     }
   }
 
@@ -175,10 +178,16 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
     }
   }
 
-  it should "publish batch of SendMessageRequests and pull them" taggedAs Integration in {
+  it should "publish batch of SendMessageRequests without individual delays and pull them" taggedAs Integration in {
     new IntegrationFixture {
       // #batch-send-request
-      val messages = for (i <- 0 until 10) yield SendMessageRequest.builder().messageBody(s"Message - $i").build()
+      val messages = for (i <- 0 until 10)
+        yield
+          SendMessageRequest
+            .builder()
+            .messageBody(s"Message - $i")
+            .delaySeconds(0)
+            .build()
 
       val future = Source
         .single(messages)
@@ -188,6 +197,28 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
       future.futureValue shouldBe Done
 
       receiveMessages(10) should have size 10
+    }
+  }
+
+  it should "publish batch of SendMessageRequests with individual delays and pull them" taggedAs Integration in {
+    new IntegrationFixture {
+      val messageDelaySeconds = 5
+      val messages = for (i <- 0 until 10)
+        yield
+          SendMessageRequest
+            .builder()
+            .messageBody(s"Message - $i")
+            .delaySeconds(messageDelaySeconds) // SQS queue should process messages after 5 seconds
+            .build()
+
+      val future = Source
+        .single(messages)
+        .runWith(SqsPublishSink.batchedMessageSink(queueUrl))
+
+      future.futureValue shouldBe Done
+
+      receiveMessages(10) should have size 0 // delay hasn't elapsed
+      receiveMessages(10, messageDelaySeconds + 5) should have size 10 // delay has elapsed
     }
   }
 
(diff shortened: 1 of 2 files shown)


llazanja:
I haven't found any existing paragraph in the docs for sqs which should be updated in regard to these changes. Let me know if you think that we need a separate section for this or another note in the existing section "Group messages and publish batches to an SQS queue".

Also, note that both `SqsPublishFlow` and `SqsPublishSink` objects use the `SqsPublishFlow.batch` method. In the integration tests, I 
- updated one which uses a sink (with no delays)
- created a new test which also uses a sink, but messages have delays (and are thus not pulled from the sqs queue)

In my opinion this should be enough to prove that delays are propagated to the `sendMessageBatch` method of a `SQSClient`.

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
```diff
@@ -191,6 +197,28 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
...
 
+  it should "publish batch of SendMessageRequests with individual delays and pull empty seq" taggedAs Integration in {
+    new IntegrationFixture {
+      // #batch-send-request
+      val messages = for (i <- 0 until 10)
+        yield
+          SendMessageRequest
+            .builder()
+            .messageBody(s"Message - $i")
+            .delaySeconds(60) // SQS queue should process messages after 60 seconds
```
pjfanning, line 208: Is there any chance that this could be set to something like 5 seconds and you could check that no messages arrive within 3 seconds (5 minus some lambda) and then check if the message arrives within the following 10 seconds? These numbers mean the test won't take too long to run but should hopefully be pretty stable.

llazanja: Done, it took me a bit of time because I couldn't realize why the request to the SQS queue didn't want to wait at all.

The issue was in the `receiveMessages` method. Wait time has to be specified in the SQS client request.

llazanja: Also, the reason why I didn't want to do something like

```
      receiveMessages(10, messageDelaySeconds - 1) should have size 0 // delay hasn't elapsed
      receiveMessages(10, messageDelaySeconds + 1) should have size 10 // delay has elapsed
```

Is that both of these calls block the thread and the total amount of waiting would be 2 x messageDelaySeconds.
I wanted to make both of these method calls async, but that would require creating another `recieveMessages` function which would return a future, and that handling this in another way.

Decided to not go with that approach cause it seemed more complicated, but can do it if you think there's need to do so.

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
```diff
@@ -191,6 +198,30 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
...
+
+      val future = Source
+        .single(messages)
+        .runWith(SqsPublishSink.batchedMessageSink(queueUrl))
+      // #batch-send-request
+
+      future.futureValue shouldBe Done
+
+      receiveMessages(10) should have size 0 // delay hasn't elapsed
+      receiveMessages(10, messageDelaySeconds + 1) should have size 10 // delay has elapsed
```
pjfanning, line 221: thanks - this looks good but I would be concerned that 3 seconds + 1 might be a bit brittle on GitHub CI infrastructure

I'm happy enough to merge as is and worry about refactoring further if we hit issues with the test failing intermittently.

I think I'll leave this open for a couple of days to let other people review it.

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
```diff
@@ -191,6 +198,30 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
     }
   }
 
+  it should "publish batch of SendMessageRequests with individual delays and pull them" taggedAs Integration in {
+    new IntegrationFixture {
+      // #batch-send-request
```
Roiocam, line 203: these was paradox documentation syntax, could you remove it? because it is duplicated.
```suggestion
```

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
```diff
@@ -191,6 +198,30 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
...
+          SendMessageRequest
+            .builder()
+            .messageBody(s"Message - $i")
+            .delaySeconds(messageDelaySeconds) // SQS queue should process messages after 5 seconds
+            .build()
+
+      val future = Source
+        .single(messages)
+        .runWith(SqsPublishSink.batchedMessageSink(queueUrl))
+      // #batch-send-request
```
Roiocam, line 216: same here.
```suggestion
```

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
llazanja: @pjfanning are you concerned in general about this type of test which relies on time delays, or are you concerned about +1 specifically, in terms that it's too short?

Cause I can tune it immediatelly

pjfanning: I'm possibly overthinking this but we need to find the right balance between having a test that is robust but that doesn't take excessively long to run. GitHub CI is a great resource but the VMs can run a bit slower and less uniformly than a PC. This is why I'm often suspicious when I see artbitrary values like 1 second.
If you have any ideas, try them out.
The ideal would be that for the test of the receiving the messages after the delay that we actively gather the messages and pass as soon as we have the 10. We can set some deadline and this can be generous. The idea is that usually, the 10 messages should arrive very quickly after the delay expires but that occasionally, we might need to wait - say maybe 10 seconds, just in case. It would not be ideal to always wait the extra 10 seconds.
I haven't run your test myself (yet) but may have time tonight to run it a few times to see how it works over a few runs.

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
```diff
@@ -13,8 +13,7 @@
 
 package docs.scaladsl
 
-import java.util.concurrent.TimeUnit
-
+import java.util.concurrent.{ TimeUnit, TimeoutException }
```
pjfanning, line 16: seems like this change is not needed now

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
llazanja: I understand what you mean.

In that case, the current setup should be good enough. By setting `waitTimeSeconds` in the `ReceiveMessageRequest`, we specify the **maximum** time which can elapse when waiting for the messages during the polling process. Also, the `get` method which is used on the future, which is returned by `awsSqsClient.receiveMessage`, has a timeout on itself, which should be larger than `waitTimeSeconds`, just to ensure that we won't be waiting for the future to complete indefinitely (however, this case shouldn't happen).

So, to summarize, we don't have a way of waiting for the delay to pass, unless we specify `waitTimeSeconds` in the request. `timeoutSeconds` is then just a precaution to ensure that we will eventually stop waiting for the future to complete.

The reason why I used `messageDelaySeconds + 1` in the method call is to ensure that we always pass a number which is larger than the delay (instead of using an arbitrary value like 60, or another variable, because it would need to have a value larger than `messageDelaySeconds`.

I can increase this arbitrary number to 5 or 10, just in case there is a hiccup in the test container for sqs, but technically, there can still be cases where this doesn't help either.

Let me know if this helps / clarifies anything or should I rephrase?

llazanja: I updated the `messageDelaySeconds` to be 5 seconds, and argument which is passed to `receiveMessages` is now 10 seconds in total (`waitTimeSeconds`). Timeout of the future (`timeoutSeconds`) will be set to 12 seconds.

But, if you run the test a couple of times, you'll notice that it finishes after just 5 seconds. Reason is that `waitTimeSeconds` is the maximum time that the client will spend waiting for the messages.

sqs/src/test/scala/docs/scaladsl/SqsPublishSpec.scala
```diff
@@ -43,18 +44,20 @@ class SqsPublishSpec extends AnyFlatSpec with Matchers with DefaultTestContext w
         .asScala
         .head
 
-    def receiveMessages(maxNumberOfMessages: Int): Seq[Message] = {
+    def receiveMessages(maxNumberOfMessages: Int, waitTimeSeconds: Int = 0): Seq[Message] = {
       // see https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/model/ReceiveMessageRequest.html
       require(maxNumberOfMessages > 0 && maxNumberOfMessages <= 10, "maxNumberOfMessages must be in 1 to 10")
 
+      val timeoutSeconds = messageRecieveMinimumTimeout + waitTimeSeconds
```
djjorjinho, line 51: There's a typo in message**Recieve**MinimumTimeout.

llazanja: Updated

Also squashed commits

laglangyue:
simple review， looks good

laglangyue:
please check the fmt.

llazanja:
Sorry, I missed this part completely in the guidelines - https://github.com/apache/pekko-connectors/blob/main/CONTRIBUTING.md#applying-code-style-to-the-project

I applied the scalafmt code style now and ran the check again, it finished successfully.

pjfanning (approved):
lgtm - thanks @llazanja

PR #764, opened 2024-08-04
fix docker-compose CI issues

pjfanning:
see https://github.com/orgs/community/discussions/116610

version param in yml is also deprecated

PR #624, opened 2024-04-21
Support Couchbase3

laglangyue:
# motivation
relates to https://github.com/apache/pekko-connectors/issues/623
support couchbase3
should be after from dropping couchbase2
# test
have some test, not finished
# docs
not finished

PR #765, opened 2024-08-04
docker-compose changes (#764)

pjfanning:
cherry pick 7f8736a6c96c54668038434b2c524ab15360ad2c #764

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,44 +25,51 @@ services:
     ports:
       - "9042:9042"
   couchbase:
-    image: couchbase:community-5.1.1
+    image: couchbase:community-7.6.1
     ports:
-      - "8091-8094:8091-8094"
-      - "11210:11210"
+      - "8091-8096:8091-8096"
+      - "11210-11211:11210-11211"
   couchbase_prep:
-    image: couchbase:community-5.1.1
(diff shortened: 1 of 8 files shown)


docker-compose.yml
```diff
@@ -26,44 +25,51 @@ services:
     ports:
       - "9042:9042"
   couchbase:
-    image: couchbase:community-5.1.1
+    image: couchbase:community-7.6.1
     ports:
-      - "8091-8094:8091-8094"
-      - "11210:11210"
+      - "8091-8096:8091-8096"
+      - "11210-11211:11210-11211"
```
Roiocam, line 31: didn't notice this on the previous PR, Is it verified?

Roiocam (approved):

docker-compose.yml
pjfanning: * this change is from #624 
* the couchbase tests pass when I copy back this change but they fail if I leave the values as is
* the tests are not run on the 1.0.x branch unless we have a PR and unfortunately the GitHub CI changes to remove support for Docker v1 require us to update some of our docker-compose configs
* It's possible that the port changes are not all needed but I thought it was better to not customise the changes and just to try to keep the docker config that works in the main branch and use it in the 1.0.x branch too

PR #783, opened 2024-09-03
try latest google-cloud-pubsub

pjfanning:
Had been set up to not be updated but we've sorted out our protobuf and grpc-java issues so this should be ok to upgrade now. I will do a separate PR for the remaining pinned google jar.


pjfanning:
will merge this into #784

PR #784, opened 2024-09-03
proto-google-cloud-bigquerystorage-v1 and google-cloud-pubsub upgrade

pjfanning:
Had been set up to not be updated but we've sorted out our protobuf and grpc-java issues so this should be ok to upgrade now.
Upgrades proto-google-cloud-bigquerystorage-v1 and google-cloud-pubsub.


pjfanning:
Would anyone have time to review this?

PR #799, opened 2024-09-12
Improve error message in case of response failure

mdedetrich:
Currently if you get a 4xx/5xx error code (i.e. insufficient permissions, bucket doesn't exist etc etc), the current logic tries to get the `Location` header (which obviously won't exist in the case of a failure) and throws an unhelpful "No Location header"`.

This change makes it so that if you do get such a failure status code, it will print out both the status code and the body. There is also a minor performance improvement with usage of `ExecutionContexts.paraistic`.


google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
```diff
@@ -93,10 +94,16 @@ private[connectors] object ResumableUpload {
...
-        }
+      Unmarshaller.withMaterializer { _ => implicit mat => (response: HttpResponse) =>
+        if (response.status.isSuccess())
+          response.discardEntityBytes().future.map { _ =>
+            response.header[Location].fold(throw InvalidResponseException(ErrorInfo("No Location header")))(_.uri)
+          }(ExecutionContexts.parasitic)
+        else
+          response.entity.dataBytes.runWith(Sink.fold(ByteString.empty)(_ ++ _)).flatMap { body =>
+            Future.failed(InvalidResponseException(
+              ErrorInfo(s"Bad response, status code: ${response.status.intValue()} and body: $body")))
```
pjfanning, line 105: Can we make this configurable? I've worked with companies in the past who can be very wary about having data appear in logs.

mdedetrich: Its the exact same as https://github.com/apache/pekko-connectors/blob/26fa1742887489828c5e0d46f1ed10f7ac056fc9/google-cloud-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/GCStorageStream.scala#L166-L168 , in fact I believe the intention was for that code to trigger in this case but due to how the order of http requests this didn't happen.

The data appearing in the logs is coming from google cloud storage, not anything that would be in the user code.

raboof: While this is client code, it's plausible that this client would be used in some web application, with the exception bubbling up to the server-side exception handling.

It might be unlikely that the server would be user-controllable or pointing to an untrusted server, exposing the response body could still leak sensitive information. I think it'd be sensible to move that to the `details` in both cases.

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
```diff
@@ -93,10 +94,16 @@ private[connectors] object ResumableUpload {
...
-        }
+      Unmarshaller.withMaterializer { _ => implicit mat => (response: HttpResponse) =>
+        if (response.status.isSuccess())
+          response.discardEntityBytes().future.map { _ =>
+            response.header[Location].fold(throw InvalidResponseException(ErrorInfo("No Location header")))(_.uri)
+          }(ExecutionContexts.parasitic)
+        else
+          Unmarshal(response.entity).to[String].flatMap { errorString =>
+            Future.failed(InvalidResponseException(
+              ErrorInfo(s"Resumable upload failed with status ${response.status}: $errorString")))
```
raboof, line 105: responding with the status code instead of a misleading `No Location header` is a great improvement.

raboof (commented):
Great improvement. Moving the response body to the `details` field would be good though imo.

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
mdedetrich: My only pushback here is that without the status code and/or body (primarily body), its not really that much more helpful then the current case. As an alternative I can put more work into the PR to figure out which status codes/body responses point to which types of errors but the [gcs resumable upload docs](https://cloud.google.com/storage/docs/performing-resumable-uploads) are really badly documented and don't specify the type of errors.

Also if we are subscribing to this strategy (which is fine) there are actually other places in the code where we can do the same type of leaking i.e. originally referenced https://github.com/apache/pekko-connectors/blob/26fa1742887489828c5e0d46f1ed10f7ac056fc9/google-cloud-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/GCStorageStream.scala#L166-L168.

Maybe @pjfanning suggestion is not a bad one, i.e. make it configurable?

raboof: I agree having the status code in the error message is fine (and I do think it's a lot less misleading than the 'missing header' error).

For the case where the error bubbles up to a Pekko HTTP endpoint it's kind of already configurable with `pekko.http.routing.verbose-error-messages` or by configuring the HTTP error handling. I guess we could consider also logging the body here, so that can be enabled/disabled with the log configuration?

Actually unmarshalling the error response to safely get the 'real' error is kinda neat but sounds like overkill.

raboof (approved):

mdedetrich:
So I have just updated PR so only status code is shown, will see how useful this is in production and if we do actually need the body contents in some way I will make a future PR.

PR #802, opened 2024-09-14
Fix error message for ResumableUpload

mdedetrich:
In https://github.com/apache/pekko-connectors/pull/799 the erorr message was improved, and i can report that it is actually working with my original expectations, and by that I mean that even the `detail` part of the error message is being reported (@raboof was right, the printing of the `detail` part is a configuration setting and its looking like its using sane default values).

As an example, right now the error message is looking like

```
java.util.concurrent.CompletionException: org.apache.pekko.stream.connectors.google.ResumableUpload$InvalidResponseException: Resumable upload failed with status: 403 Forbidden: body: {"error":{"code":403,"message":"<REDACTED> does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).","errors":[{"message":"<REDACTED> does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).","domain":"global","reason":"forbidden"}]}}
```

The only minor issue currently is that the printing of the error message has parts that are a bit duplicated so this PR fixes that. With this change the error message would look like this

```
java.util.concurrent.CompletionException: org.apache.pekko.stream.connectors.google.ResumableUpload$InvalidResponseException: Resumable upload failed with status 403 Forbidden: {"error":{"code":403,"message":"<REDACTED> does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).","errors":[{"message":"<REDACTED> does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).","domain":"global","reason":"forbidden"}]}}
```

### DIFF
--- a/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
+++ b/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
@@ -102,7 +102,7 @@ private[connectors] object ResumableUpload {
         else
           Unmarshal(response.entity).to[String].flatMap { errorString =>
             Future.failed(InvalidResponseException(
-              ErrorInfo(s"Resumable upload failed with status: ${response.status}", s"body: $errorString")))
+              ErrorInfo(s"Resumable upload failed with status ${response.status}", errorString)))
           }(ExecutionContexts.parasitic)
       }.withDefaultRetry
 


gmethvin (approved):

pjfanning (approved):
lgtm

PR #808, opened 2024-09-16
don't use 'current' for doc links

pjfanning:
* the main branch is intended for 1.1 release and is best to direct specifically to 1.1 docs of other pekko modules
* I intend to backport this to 1.0 branch because that branch needs to direct towards 1.0 versions of other pekko module docs

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -392,7 +392,7 @@ lazy val docs = project
       "hadoop.version" -> Dependencies.HadoopVersion,
       "extref.github.base_url" -> s"https://github.com/apache/pekko-connectors/tree/${if (isSnapshot.value) "main"
         else "v" + version.value}/%s",
-      "extref.pekko.base_url" -> s"https://pekko.apache.org/docs/pekko/current/%s",
+      "extref.pekko.base_url" -> s"https://pekko.apache.org/docs/pekko/${Dependencies.PekkoBinaryVersion}/%s",
       "scaladoc.org.apache.pekko.base_url" -> s"https://pekko.apache.org/api/pekko/${Dependencies.PekkoBinaryVersion}",
       "javadoc.org.apache.pekko.base_url" -> s"https://pekko.apache.org/japi/pekko/${Dependencies.PekkoBinaryVersion}/",
       "javadoc.org.apache.pekko.link_style" -> "direct",
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -21,7 +21,7 @@ object Dependencies {
   val ScalaVersions = Seq(Scala213, Scala212, Scala3)
 
   val PekkoVersion = PekkoCoreDependency.version
-  val PekkoBinaryVersion = PekkoVersion.take(3)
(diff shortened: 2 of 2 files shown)


Roiocam (approved):

build.sbt
```diff
@@ -392,7 +392,7 @@ lazy val docs = project
       "hadoop.version" -> Dependencies.HadoopVersion,
       "extref.github.base_url" -> s"https://github.com/apache/pekko-connectors/tree/${if (isSnapshot.value) "main"
         else "v" + version.value}/%s",
-      "extref.pekko.base_url" -> s"https://pekko.apache.org/docs/pekko/current/%s",
+      "extref.pekko.base_url" -> s"https://pekko.apache.org/docs/pekko/${Dependencies.PekkoBinaryVersion}/%s",
```
raboof, line 395: Now that I think about it more, I'm no longer sure that it's better to link to the the Pekko version this is built against than the 'current' Pekko version: because of binary compatibility it'll be common and recommended that people use newer versions of Pekko with older versions of satellite libraries.

I'm not opposed though :)

project/Dependencies.scala
```diff
@@ -32,7 +32,7 @@ object Dependencies {
   val PekkoGrpcBinaryVersion = "1.1"
   val PekkoHttpVersion = PekkoHttpDependency.version
   val PekkoStreamsCirceVersion = "1.1.0"
-  val PekkoHttpBinaryVersion = "1.0"
+  val PekkoHttpBinaryVersion = PekkoHttpVersion.take(3)
```
raboof, line 35: Maybe

```
import sbt.librarymanagement.VersionNumber

VersionNumber(PekkoHttpVersion).take(2).mkString(".")
```
?

build.sbt
pjfanning: The idea is to point to the docs of the lib based on the version of that lib that we build with. As we have seen with the Pekko core docs and the redirect of current from 1.0 to 1.1 broke some links. This approach of not linking to current means things are more stable.

raboof: > As we have seen with the Pekko core docs and the redirect of current from 1.0 to 1.1 broke some links

I think we should avoid that in the future by leaving behind redirects, though: while changing our own sites to link to specific version indeed avoids some breakage, there might still be external links pointing to 'current'. It might also be better for SEO. Additionally, in Akka we often noticed people being confused because they were referring to outdated versions of the docs.

Having said that, I'm still OK with the change to the versioned docs - just wanted to share my thoughts for consideration.

pjfanning: I've already pushed similar changes into other Pekko repos so I would like to stay consistent. We can always review the strategy again later.

The Akka docs have a feature to warn you if you are viewing old docs and provides a link to the latest docs. We could try to do something similar. One approach would be for legacy branches (eg 1.0.x branches) to be modified to put a header on all pages that provides a link to the `current` version of that page.

raboof (approved):

PR #809, opened 2024-09-17
Set current version to latest 1.0.3

mdedetrich:
Currently the Pekko artifacts/snapshots are broken, here are the stats from today

```
latest = 1.1.0-M1+65-53ed673e-SNAPSHOT (not working)
2 days = 1.1.0-M1+64-600516f4-SNAPSHOT (not working)
3 days = 1.1.0-M1+58-20bcdd5f-SNAPSHOT (working)
4 days = 1.1.0-M1+57-eb3ae265-SNAPSHOT (working)
one week = 1.1.0-M1+56-67d9adfa-SNAPSHOT (working)
```

The commit which broke the changes is https://github.com/apache/pekko-connectors/pull/801. The issue here is if you set `currentVersion` to some version, that is actually setting the minimum bound. More specifically, at work I had my pekko core version still set to 1.1.0 but if I set pekko-connectors to a not working snapshot (lets say `1.1.0-M1+65-53ed673e-SNAPSHOT`) then it would fail with class not found errors.

This is because Pekko is doing a check on the major/minor version, so we have to use the latest 1.0.x version.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -22,6 +22,7 @@ object Dependencies {
 
   val PekkoVersion = PekkoCoreDependency.version
   val PekkoBinaryVersion = PekkoVersion.take(3)
+  val PekkoConnectorsKafkaVersion = "1.0.0"
 
   val InfluxDBJavaVersion = "2.23"
 
@@ -189,7 +190,7 @@ object Dependencies {
     libraryDependencies ++= Seq(
       "org.apache.pekko" %% "pekko-slf4j" % PekkoVersion,
       "org.apache.pekko" %% "pekko-stream-testkit" % PekkoVersion % Test,
-      "org.apache.pekko" %% "pekko-connectors-kafka" % "1.1.0-M1" % Test,
+      "org.apache.pekko" %% "pekko-connectors-kafka" % PekkoConnectorsKafkaVersion % Test,
       "junit" % "junit" % "4.13.2" % Test,
       "org.scalatest" %% "scalatest" % ScalaTestVersion % Test))
 
--- a/project/PekkoCoreDependency.scala
+++ b/project/PekkoCoreDependency.scala
@@ -20,5 +20,5 @@ import com.github.pjfanning.pekkobuild.PekkoDependency
 object PekkoCoreDependency extends PekkoDependency {
   override val checkProject: String = "pekko-cluster-sharding-typed"
   override val module: Option[String] = None
-  override val currentVersion: String = "1.1.1"
+  override val currentVersion: String = "1.0.3"
 }
--- a/project/PekkoHttpDependency.scala
+++ b/project/PekkoHttpDependency.scala
@@ -20,5 +20,5 @@ import com.github.pjfanning.pekkobuild.PekkoDependency
 object PekkoHttpDependency extends PekkoDependency {
   override val checkProject: String = "pekko-http-testkit"
   override val module: Option[String] = Some("http")
-  override val currentVersion: String = "1.1.0-M1"
+  override val currentVersion: String = "1.0.1"
 }


raboof:
> then it would fail with class not found errors.
> This is because Pekko is doing a check on the major/minor version

I'm confused: is it failing with class not found errors or failing the major/minor version check? Can you share a non-working project?

> if you set `currentVersion` to some version, that is actually setting the minimum bound

I thought it was intended that `main` would be targeting `1.1.x`, so setting that as the minimum bound seems valid?

mdedetrich:
> > then it would fail with class not found errors.
> > This is because Pekko is doing a check on the major/minor version
> 
> I'm confused: is it failing with class not found errors or failing the major/minor version check? Can you share a non-working project?
> 
> > if you set `currentVersion` to some version, that is actually setting the minimum bound
> 
> I thought it was intended that `main` would be targeting `1.1.x`, so setting that as the minimum bound seems valid?

In my work project I have this

```kotlin
val pekkoVersion = "1.1.0"
val pekkoConnectorsVersion = "1.1.0-M1+65-53ed673e-SNAPSHOT"
```

where pekkoVersion is pointing to the pekko-core/pekko-streams version

if we keep pekkoVersion constant to 1.1.0 then the pekkoConnectorsVersion starts breaking at `1.1.0-M1+64-600516f4-SNAPSHOT` and the reason why is this commit https://github.com/apache/pekko-connectors/pull/801. It seems like there is some code within pekko core that checks the major/minor/patch and because `currentVersion = 1.1.1` is greater than `pekkoVersion = 1.1.0` its causing `ClassNotFound` errors for some reason

raboof:
> In my work project I have this
> 
> ```kotlin
> val pekkoVersion = "1.1.0"
> val pekkoConnectorsVersion = "1.1.0-M1+65-53ed673e-SNAPSHOT"
> ```
> 
> where pekkoVersion is pointing to the pekko-core/pekko-streams version
> 
> if we keep pekkoVersion constant to 1.1.0 then the pekkoConnectorsVersion starts breaking at `1.1.0-M1+64-600516f4-SNAPSHOT` and the reason why is this commit #801. It seems like there is some code within pekko core that checks the major/minor/patch and because `currentVersion = 1.1.1` is greater than `pekkoVersion = 1.1.0` its causing `ClassNotFound` errors for some reason

It's difficult to say without a reproducer or exact error message, but you might be referring to https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed . Are you seeing that message? If so the solution is to pull up the Pekko version in your project for all your (possibly-transitive) dependencies, as described there.

pjfanning:
We are trying to do releases. And one aspect of the release is that Scala 2 inlining is enabled. This requires that we build with 1.1 jars from our various modules.

I have noticed in the past that the Pekko core Pekko version checks do seem to be hard to work with when you have Pekko snapshots in your dependencies.

You can set the config `pekko.fail-mixed-versions=off` to disable the check.

Can we spend time trying to fix the actual Pekko checks as opposed to just reverting all the build upgrades - which will throw the whole 1.1 release process off track?

mdedetrich:
What specifically fails is `ActorSystem.create()` which throws an `java.lang.IllegalStateException`, I think this is the relevant code https://github.com/apache/pekko/blob/c839a2b85837024eb335bd5fb5b3eff920beabfd/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala#L843-L850

raboof:
> If we use one of the "non working" versions such as 1.1.0-M1+65-53ed673e-SNAPSHOT but bump pekkVersion to 1.1.1 then it works again, but the whole point is we are meant to be using SemVer and so the project should work even if pekkoVersion is set to 1.0.0

I'm OK with keeping the minimum pekko version in pekko-connectors on 1.0.3, but I thought we wanted pekko-connectors 1.1.x to require pekko 1.1.x to get the benefits of inlining?

mdedetrich:
> I'm OK with keeping the minimum pekko version in pekko-connectors on 1.0.3, but I thought we wanted pekko-connectors 1.1.x to require pekko 1.1.x to get the benefits of inlining?

This is true, at worst we may have to get rid of all of the inlining but lets diagnose whats really going on herre.

pjfanning:
> I think @pjfanning is right in that we probably need to update the check. We do need to cancel the current `-RC`'s for now since the jar's seem to be broken

I won't release anything until we find the root cause of the version check issues. I'll leave the votes open but will cancel them if we find that we need big changes. In practice, we could release them and release note that users need to use Pekko 1.1.1 until we sort out this issue.

mdedetrich:
So this is the exact error that is being thrown in my project

```
2024-09-17 11:21:59.865 [Test worker] ERROR <REDACTED> - Actor System error
java.lang.IllegalStateException: You are using version 1.1.1 of Apache Pekko, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. You can solve this by adding an explicit dependency on version 1.1.1 of the [pekko-stream-testkit, pekko-slf4j, pekko-testkit, pekko-actor-typed, pekko-actor-testkit-typed] artifacts to your project. Here's a complete collection of detected artifacts: (1.1.0, [pekko-actor-testkit-typed, pekko-actor-typed, pekko-slf4j, pekko-stream-testkit, pekko-testkit]), (1.1.1, [pekko-actor, pekko-protobuf-v3, pekko-stream]). See also: https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed
	at org.apache.pekko.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:188)
	at org.apache.pekko.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:166)
	at org.apache.pekko.actor.ActorSystemImpl.liftedTree2$1(ActorSystem.scala:1051)
	at org.apache.pekko.actor.ActorSystemImpl._start$lzycompute(ActorSystem.scala:1040)
	at org.apache.pekko.actor.ActorSystemImpl._start(ActorSystem.scala:1039)
	at org.apache.pekko.actor.ActorSystemImpl.start(ActorSystem.scala:1063)
	at org.apache.pekko.actor.ActorSystem$.apply(ActorSystem.scala:282)
	at org.apache.pekko.actor.ActorSystem$.apply(ActorSystem.scala:326)
	at org.apache.pekko.actor.ActorSystem.create(ActorSystem.scala:270)
```

I think that this may be a false alarm. It is because the pekko snapshot suddenly updated to 1.1.1 while I was using 1.1.0 and due to transitive resolution it brought in mixed versions of Pekko

pjfanning:
> So this is the exact error that is being thrown in my project
> 
> ```
> 2024-09-17 11:21:59.865 [Test worker] ERROR <REDACTED> - Actor System error
> java.lang.IllegalStateException: You are using version 1.1.1 of Apache Pekko, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. You can solve this by adding an explicit dependency on version 1.1.1 of the [pekko-stream-testkit, pekko-slf4j, pekko-testkit, pekko-actor-typed, pekko-actor-testkit-typed] artifacts to your project. Here's a complete collection of detected artifacts: (1.1.0, [pekko-actor-testkit-typed, pekko-actor-typed, pekko-slf4j, pekko-stream-testkit, pekko-testkit]), (1.1.1, [pekko-actor, pekko-protobuf-v3, pekko-stream]). See also: https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed
> 	at org.apache.pekko.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:188)
> 	at org.apache.pekko.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:166)
> 	at org.apache.pekko.actor.ActorSystemImpl.liftedTree2$1(ActorSystem.scala:1051)
> 	at org.apache.pekko.actor.ActorSystemImpl._start$lzycompute(ActorSystem.scala:1040)
> 	at org.apache.pekko.actor.ActorSystemImpl._start(ActorSystem.scala:1039)
> 	at org.apache.pekko.actor.ActorSystemImpl.start(ActorSystem.scala:1063)
> 	at org.apache.pekko.actor.ActorSystem$.apply(ActorSystem.scala:282)
> 	at org.apache.pekko.actor.ActorSystem$.apply(ActorSystem.scala:326)
> 	at org.apache.pekko.actor.ActorSystem.create(ActorSystem.scala:270)
> ```
> 
> I think that this may be a false alarm. It is because the pekko snapshot suddenly updated to 1.1.1 while I was using 1.1.0 and due to transitive resolution it brought in mixed versions of Pekko

I have found this in the past - that you end up having to add a lot more explicit dependencies to fight against old transitive dependencies.
Or you could add the pekko-bom to your dependencies. 

https://mvnrepository.com/artifact/org.apache.pekko/pekko-bom

mdedetrich:
Closing this PR, I will also cancel my -1 vote in the releases

PR #832, opened 2024-09-26
configure aws-spi-pekko-http integration tests in nightly builds

jtjeferreira:
As discussed in https://github.com/apache/pekko-connectors/pull/827#issuecomment-2375129573, this PR configures `aws-spi-pekko-http` integration tests.

I understand that S3 module integration tests are setup in a different way (using `@DoNotDiscover` and explicitly calling `org.scalatest.tools.Runner`) , but I don't want to touch for now.

I changed the way credentials are passed to use environment variables. I did this, because for the `aws-spi-pekko-http` tests the credentials cannot be configured using JVM system properties, but both tests support environment variables as part of the default credential provider.

### DIFF
--- a/.github/workflows/nightly-builds.yaml
+++ b/.github/workflows/nightly-builds.yaml
@@ -56,12 +56,14 @@ jobs:
       - name: Cache Coursier cache
         uses: coursier/cache-action@v6
 
-      - name: S3 Integration tests
+      - name: AWS connectors Integration tests
+        env:
+          AWS_ACCESS_KEY: ${{ secrets.AWS_ACCESS_KEY }}
+          AWS_SECRET_KEY: ${{ secrets.AWS_SECRET_KEY }}
         run:  |-
           sbt \
-          -Dpekko.connectors.s3.aws.credentials.provider=static \
-          -Dpekko.connectors.s3.aws.credentials.access-key-id=${{ secrets.AWS_ACCESS_KEY }} \
-          -Dpekko.connectors.s3.aws.credentials.secret-access-key=${{ secrets.AWS_SECRET_KEY }} \
           -Dpekko.stream.connectors.s3.scaladsl.AWSS3IntegrationSpec.enableListAllMyBucketsTests=false \
           -Dpekko.stream.connectors.s3.scaladsl.AWSS3IntegrationSpec.enableMFATests=false \
           + "s3/Test/runMain org.scalatest.tools.Runner -o -s org.apache.pekko.stream.connectors.s3.scaladsl.AWSS3IntegrationSpec"
+          sbt \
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-builds.yaml
```diff
@@ -56,12 +57,14 @@ jobs:
       - name: Cache Coursier cache
         uses: coursier/cache-action@v6
 
-      - name: S3 Integration tests
+      - name: AWS connectors Integration tests
+        env:
+          AWS_ACCESS_KEY: ${{ secrets.AWS_ACCESS_KEY }}
```
jtjeferreira, line 62:
```suggestion
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY }}
```

jtjeferreira:
@pjfanning thanks for running the workflows. For some reason the secret is not being properly propagated. 

If I look into https://github.com/apache/pekko-connectors/actions/runs/11044846930/job/30681466278, I see that the secret is masked as expected: 
```
sbt \
  -Dpekko.connectors.s3.aws.credentials.provider=static \
  -Dpekko.connectors.s3.aws.credentials.access-key-id=*** \
```

However in https://github.com/apache/pekko-connectors/actions/runs/11049659282/job/30695897761?pr=832 it is empty: 
```
env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/8.0.422-5/x64
    JAVA_HOME_8_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/8.0.422-5/x64
    AWS_ACCESS_KEY_ID: 
```

I don't have much experience with github actions...

pjfanning:
It looks like there are possibly some restrictions set up so that the access key secret can not be used as an env variable. I'll raise an issue with the ASF Infra team to see if they can give some guidance. This could take a while.
Can you add an 'assume' section to any tests that need the env variables and have the assume section check if the env variable is non-empty? This would mean that the test is skipped if the env variable is not there.

jtjeferreira:
@pjfanning I reverted all my changes to make sure the secrets can be accessed in a PR. If it works, I will progress from there...

jtjeferreira:
> @pjfanning I reverted all my changes to make sure the secrets can be accessed in a PR. If it works, I will progress from there...

As I suspected [the secrets cannot be accessed in a PR](https://github.com/apache/pekko-connectors/actions/runs/11050669781/job/30699348927?pr=832). Can that be changed?

pjfanning:
The ASF Infra team answered and said that GitHub forks of the main project don't have access to the secrets.

In theory, I could create a branch on the main project and create a PR off of it. Let me try that.

PR #861, opened 2024-10-09
v1.1.0 release notes

pjfanning:
release under discussion

### DIFF
--- a/.github/workflows/publish-1.1-docs.yml
+++ b/.github/workflows/publish-1.1-docs.yml
@@ -26,17 +26,17 @@ jobs:
 
       - name: Build Documentation
         run: |-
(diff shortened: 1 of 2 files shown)


JD557:
Should the release notes have an extra note regarding https://github.com/apache/pekko-connectors/pull/314, as it is a breaking change?

At least from the wording in the changelog (Remove duplicate scopes value in reference.conf) it was not entirely obvious to me that this could break applications that were loading the scope from the `service-account.scopes`.

pjfanning:
@mdedetrich what do you think about this being a breaking change?

I guess we can move this into its own 'Potentially Breaking Changes' section and try to provide more details about who might see issues.

He-Pin (approved):
lgtm

PR #943, opened 2025-01-10
increase build memory

pjfanning:
running into issues trying to build a release on my laptop


pjfanning:
* looks like SoftReferenceCacheSpec relies on constrained memory
* I'll see if there is another way to increase memory while publishing

pjfanning:
I reduced the memory a bit. I think the extra memory will help but now we don't get CI test issues.

raboof (approved):

PR #949, opened 2025-01-12
Slick - Add slick flow functions with  error handling using try.

minjibir:


pjfanning:
My preference would be not to change the 'Slick' object. Could you just create just create a 2nd 'SlickWithTryResult' or something like that?

minjibir:
Absolutely. That's my first implementation, but later settled with this. I will change it.

pjfanning (changes requested):
with both new files, they are based on existing files and they should use the source headers from those files - so can you replace the source headers in your new files with

```
/*
 * Licensed to the Apache Software Foundation (ASF) under one or more
 * license agreements; and to You under the Apache License, version 2.0:
 *
 *   https://www.apache.org/licenses/LICENSE-2.0
 *
 * This file is part of the Apache Pekko project, which was derived from Akka.
 */

/*
 * Copyright (C) since 2016 Lightbend Inc. <https://www.lightbend.com>
 */
```

minjibir:
@pjfanning file headers updated. Thank you for pointing that out.

pjfanning:
the code doesn't compile in scala 2.12

pjfanning:
I have committed some code changes but I'm not sure if they work yet.

slick/src/test/scala/docs/scaladsl/SlickWithTryResultSpec.scala
```diff
@@ -0,0 +1,355 @@
...
+  "SlickSession.forDbAndProfile" must {
+    "create a slick session able to talk to the db" in {
+      // #init-session-from-db-and-profile
+      val db = Database.forConfig("slick-h2.db")
+      val profile = slick.jdbc.H2Profile
+      val slickSessionCreatedForDbAndProfile: SlickSession = SlickSession.forDbAndProfile(db, profile)
+      // #init-session-from-db-and-profile
+      try {
+        val q = sql"select true".as[Boolean]
+        val result = Slick
```
pjfanning, line 105: @minjibir why is this testing the 'Slick' object and not the new SlickWithTryResult object?

minjibir: Let me check it out.

minjibir: I was trying to keep it DRY by using the source from `Slick` to setup the test since the implimentation is identical, and that the main purpose of the `SlickWithTryResult` is during the write operations. Would it be better if I have the same `source` function in `SlickWithTryResult` as well?

pjfanning: why not delete this test case if the test is already in the SlickSpec?

minjibir: Ok, I will do so.

pjfanning (approved):
thanks - lgtm

PR #969, opened 2025-01-29
CI: restrict number of parallel jobs

pjfanning:
GitHub CI have requested generally that ASF projects avoid running more than 15 jobs in parallel. This CI jobs has a few other jobs as well as the matrix jobs, that's why I choose 10 as the limit here.


nvollmar (approved):
lgtm

PR #981, opened 2025-02-09
test: avoid topic overlap between Mqtt tests

raboof:
Avoid multiple tests using the same topic.

Possibly towards avoiding #468


raboof:
it seems to consistently fail with this change? very surprising, but would be cool if that gives us a lead ;)

PR #1002, opened 2025-03-11
FILE: Add size and date information to ZipArchiveMetadata

jypma:
This brings the functionality closer to TarArchiveMetadata that already has these. Size and date information can be very valuable when unzipping files, to check if this entry has already been unzipped.

### DIFF
--- a/file/src/main/scala/org/apache/pekko/stream/connectors/file/model.scala
+++ b/file/src/main/scala/org/apache/pekko/stream/connectors/file/model.scala
@@ -25,11 +25,16 @@ object ArchiveMetadata {
   def create(filePath: String): ArchiveMetadata = new ArchiveMetadata(filePath)
 }
 
-final case class ZipArchiveMetadata(name: String) {
+final case class ZipArchiveMetadata(name: String)(val size: Long, val lastModification: Instant) {
   def getName() = name
 }
 object ZipArchiveMetadata {
-  def create(name: String): ZipArchiveMetadata = ZipArchiveMetadata(name)
+  @deprecated("Use the variant with 3 arguments instead.", since = "pekko-connectors-file:1.1.0")
+  def apply(name: String): ZipArchiveMetadata = new ZipArchiveMetadata(name)(0, Instant.EPOCH)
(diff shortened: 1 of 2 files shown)


file/src/main/scala/org/apache/pekko/stream/connectors/file/model.scala
```diff
@@ -25,11 +25,13 @@ object ArchiveMetadata {
   def create(filePath: String): ArchiveMetadata = new ArchiveMetadata(filePath)
 }
 
-final case class ZipArchiveMetadata(name: String) {
+final case class ZipArchiveMetadata(name: String, size: Long, lastModification: Instant) {
```
raboof, line 28: this seems like it'd not be binary compatible, because of the generated methods for case classes that would change - I'm not sure if it's possible to manually 'add back' the original methods to restore binary compatibility?

raboof (commented):
seems neat to have this available

file/src/main/scala/org/apache/pekko/stream/connectors/file/model.scala
jypma: Ah, I missed that. Sure, I'll try to toss in some additional `apply` methods.

jypma: Should we tag those `deprecated`, since we'd have to set `size` and `lastModification` to zero? I thought about that for the old `create()` method as well.

raboof: I don't have a strong preference, so up to you as far as I'm concerned!

file/src/main/scala/org/apache/pekko/stream/connectors/file/model.scala
```diff
@@ -25,11 +25,16 @@ object ArchiveMetadata {
   def create(filePath: String): ArchiveMetadata = new ArchiveMetadata(filePath)
 }
 
-final case class ZipArchiveMetadata(name: String) {
+final case class ZipArchiveMetadata(name: String)(val size: Long, val lastModification: Instant) {
```
pjfanning, line 28: one param list please - not 2 - to match TarArchiveMetadata

PR #1004, opened 2025-03-15
Missing death watch in UDP.bindFlow(...)

sdeigm:
Added the death watch for the UDP listener actor. Stage is terminated with a failure when UDP listener terminates

### DIFF
--- a/udp/src/main/scala/org/apache/pekko/stream/connectors/udp/impl/UdpBind.scala
+++ b/udp/src/main/scala/org/apache/pekko/stream/connectors/udp/impl/UdpBind.scala
@@ -16,7 +16,7 @@ package org.apache.pekko.stream.connectors.udp.impl
 import java.net.InetSocketAddress
 
 import org.apache.pekko
-import pekko.actor.{ ActorRef, ActorSystem }
+import pekko.actor.{ ActorRef, ActorSystem, Terminated }
 import pekko.annotation.InternalApi
 import pekko.io.{ IO, Udp }
 import pekko.io.Inet.SocketOption
@@ -53,6 +53,7 @@ import scala.concurrent.{ Future, Promise }
     case (sender, Udp.Bound(boundAddress)) =>
       boundPromise.success(boundAddress)
       listener = sender
+      stageActor.watch(listener)
       pull(in)
     case (_, Udp.CommandFailed(cmd: Udp.Bind)) =>
       val ex = new IllegalArgumentException(s"Unable to bind to [${cmd.localAddress}]")
@@ -62,6 +63,9 @@ import scala.concurrent.{ Future, Promise }
       if (isAvailable(out)) {
         push(out, Datagram(data, sender))
       }
+    case (_, Terminated(_)) =>
+      listener = null
+      failStage(new IllegalStateException("UDP listener terminated unexpectedly"))
     case _ =>
   }
 


mdedetrich:
@sdeigm Would it be possible to add a test for this?

udp/src/main/scala/org/apache/pekko/stream/connectors/udp/impl/UdpBind.scala
```diff
@@ -62,6 +63,9 @@ import scala.concurrent.{ Future, Promise }
       if (isAvailable(out)) {
         push(out, Datagram(data, sender))
       }
+    case (_, Terminated(_)) =>
+      listener = null
+      failStage(new IllegalStateException("UDP listener terminated unexpectedly"))
```
mdedetrich, line 68: I am wondering whether a `Runtime` exception is better suited to this than an `IllegalStateException` @raboof @pjfanning wdyt?

I am also wondering if there is an already existing better exception that could be used

sdeigm:
Feel free to change the exception type, but I think its a perfect fit:

"Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation."

The UDP listener only terminates on a complete network breakdown. I my case the software is running on an edge device only connected via WLAN. The problem only happens when the device is kicked out of the WLAN due to poor reception quality.

Regarding the test: I can try to add a test. But to trigger this condition the UDP listener has to be killed. During the development I added code within the stage to kill the listener. For a standard test I have to figure out how/if I potentially can kill the listener from outside of the stage.

mdedetrich:
> The UDP listener only terminates on a complete network breakdown. I my case the software is running on an edge device only connected via WLAN. The problem only happens when the device is kicked out of the WLAN due to poor reception quality.

This really comes down to semantics on what you define as illegal. In my view a network breaking down isn't "illegal", in fact it can happen at any time legitimately and so can be considered to be part of normal program operation.

Personally I don't feel strongly about this, just wanted to get feedback from others

> Regarding the test: I can try to add a test. But to trigger this condition the UDP listener has to be killed. During the development I added code within the stage to kill the listener. For a standard test I have to figure out how/if I potentially can kill the listener from outside of the stage.

If it sounds like too much work I would timebox it, however its always ideal to have tests for cases like this.

sdeigm:
> This really comes down to semantics on what you define as illegal. In my view a network breaking down isn't "illegal", in fact it can happen at any time legitimately and so can be considered to be part of normal program operation and such its not really an "illegal state".

Sure, the network breakdown can always happen, that is not "illegal". But when the network breakdown has happened,  then "... the Java environment or Java application is not in an appropriate state for the requested operation.", it is not possible to use UDP at that point in time.

udp/src/main/scala/org/apache/pekko/stream/connectors/udp/impl/UdpBind.scala
pjfanning: ConnectionException has a good name but its Javadoc is a bit specific.

> Signals that an error occurred while attempting to connect a socket to a remote address and port. Typically, the connection was refused remotely (e.g., no process is listening on the remote address/port).

Would a plain IOException be an option?

sdeigm:
I think IOException is a good proposal, since the problem indicated is for sure IO related.

mdedetrich:
@sdeigm Did you manage to get started on the tests or is it the case that it will end up being too much work?

mdedetrich:
Oh and you also need to format the code using `scalafmt`

sdeigm:
I used ``sbt applyCodeStyle`` to update the formatting, hope that's correct.

Regarding the test: I found no way to stop the underlying UDP actor (which is not known outside of the graphstage). I tried to close the bound UDP port from outside, but that does seem to be possible from Java/Scala without the usage of JNI. So I think we must live without a test :-(

mdedetrich (approved):
lgtm

mdedetrich:
Thanks for the help, I am going to merge this now so that you can test it out when a snapshot gets published tonight.

pjfanning (commented):
lgtm

PR #1019, opened 2025-03-29
Replace ignore with DoNotDiscover and split test out

mdedetrich:
This PR replaces the ScalaTest `ignore` directive with `DoNotDiscover` which is whats used with AWS (see https://github.com/apache/pekko-connectors/blob/8c5d65051cfeb3df4a0eb87039c33131b6b21ccc/s3/src/test/scala/org/apache/pekko/stream/connectors/s3/scaladsl/S3IntegrationSpec.scala#L1281). The `DoNotDiscover` annotation means that the test won't be run when you run the general `test` command in sbt however it can be run with an explicit command as well as supporting running in Intellij directly (see comment) .

The PR also splits out the test into its implementation (the trait) and the runner (the class). The intention here is to run the integration spec against [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) (in the same way that we have S3 tests that are run against minio).


pjfanning (approved):
Lgtm

PR #1020, opened 2025-03-29
Use GoogleSettings instead of GCStorageSettings

mdedetrich:
In addition to getting rid of the depreciation message, this is necessary to add in [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) testing as only `GoogleSettings` has the necessary `settings.withCredentials(NoCredentials(projectId, ""))` method (the deprecated `GCStorageSettings` has no such method and hence it forces official google authentication which won't work with gcs-storage-settings)

I have verified this hasn't caused any regressions by testing against a real google account. Another easy way to verify is to temporarily make `ServiceAccountCredentials` public, as well as the various constructor parameters and then you can do

```scala

val credentials = settings.credentials.asInstanceOf[org.apache.pekko.stream.connectors.google.auth.ServiceAccountCredentials]
println("credentials")
println(credentials.clientEmail)
println(credentials.project)
println(credentials.privateKey)
```

And it will match the settings at https://github.com/apache/pekko-connectors/blob/1fe3474d654bf48b0e81011d878e541a32409c93/google-cloud-storage/src/test/resources/application.conf#L38-L40

which means that everything is being loaded as expected

### DIFF
--- a/google-cloud-storage/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/GoogleGCStorageStreamIntegrationSpec.scala
+++ b/google-cloud-storage/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/GoogleGCStorageStreamIntegrationSpec.scala
@@ -17,7 +17,7 @@
 
 package org.apache.pekko.stream.connectors.googlecloud.storage.impl
 
-import org.apache.pekko.stream.connectors.googlecloud.storage.GCStorageSettings
+import org.apache.pekko.stream.connectors.google.GoogleSettings
 import org.scalatest.DoNotDiscover
 
 import scala.annotation.nowarn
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/GoogleGCStorageStreamIntegrationSpec.scala
```diff
@@ -44,12 +44,9 @@ import scala.annotation.nowarn
...
 @DoNotDiscover
 class GoogleGCStorageStreamIntegrationSpec extends GCStorageStreamIntegrationSpec {
-  @nowarn("msg=deprecated")
-  def settings: GCStorageSettings = GCStorageSettings()
+  def settings: GoogleSettings = GoogleSettings()
 
   override def bucket = "connectors"
   override def rewriteBucket = "pekko-connectors-rewrite"
   override def projectId = settings.projectId
-  override def clientEmail = settings.clientEmail
```
mdedetrich, line 53: These methods aren't actually used anywhere which is why we can delete them without problems

softinio (approved):

PR #1021, opened 2025-03-30
Rename none credentials to access-token and remove token from none

mdedetrich:
This PR modifies `NoCredentials` so that you aren't forced to provide it with a token and by consequence means that no `Authorization` header will be added when using `NoCredentials`. More specifically, the constructor for `NoCredentials` has been changed so that it doesn't have a `token` parameter and the `token` value from typesafe config is no longer retrieved.

In order to get this to work, I had to split out the `get()` method (which returns the token) from `Credentials` into a separate `RetrievableCredentials` trait and every subclass of `Credentials` aside from `NoCredentials` inherits this trait.

The main reason for this change is that `NoCredentials` shouldn't, by design, add token auth headers (the official analogue from Google's own sdk [doesn't do this](https://github.com/googleapis/sdk-platform-java/blob/main/java-core/google-cloud-core/src/main/java/com/google/cloud/NoCredentials.java)), it can even be argued that its a potential security risk as it may inadvertently force user to provide a token they shouldn't have. The change is also being done in context of adding support for [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) where their own sample Java implementation also has [doesn't have to provide any token](https://github.com/fsouza/fake-gcs-server/blob/main/examples/java/src/test/java/com/fsouza/fakegcsserver/java/examples/FakeGcsServerTest.java#L25-L30)

Note that all of the touched code is dealing with internals that are marked with `DoNotInherit`/`private` so it doesn't break binary compatibility in any way. It could be argued that this is a behaviour change and so it should only land for pekko-connectors 1.2.x however an alternate view is that `NoCredentials` would have only be used for testing (I highly doubt anyone would be using in prod, but it is theoretically possible), this is ontop of the security concern mentioned earlier.

### DIFF
--- a/google-cloud-pub-sub/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/impl/PubSubApiSpec.scala
+++ b/google-cloud-pub-sub/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/impl/PubSubApiSpec.scala
@@ -16,15 +16,11 @@ package org.apache.pekko.stream.connectors.googlecloud.pubsub.impl
 import org.apache.pekko
 import pekko.Done
 import pekko.actor.ActorSystem
-import pekko.http.scaladsl.{ ConnectionContext, Http }
 import pekko.stream.connectors.googlecloud.pubsub._
 import pekko.stream.connectors.testkit.scaladsl.LogCapturing
 import pekko.stream.scaladsl.{ Keep, Sink, Source }
-import com.github.tomakehurst.wiremock.WireMockServer
 import com.github.tomakehurst.wiremock.client.WireMock
 import com.github.tomakehurst.wiremock.client.WireMock.{ aResponse, urlEqualTo }
-import com.github.tomakehurst.wiremock.common.ConsoleNotifier
-import com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig
 import com.typesafe.config.ConfigFactory
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.concurrent.ScalaFutures
@@ -34,7 +30,7 @@ import org.scalatest.matchers.should.Matchers
 import java.security.cert.X509Certificate
 import java.time.Instant
 import java.util.Base64
-import javax.net.ssl.{ SSLContext, SSLEngine, X509TrustManager }
+import javax.net.ssl.X509TrustManager
 import scala.collection.immutable.Seq
 import scala.concurrent.Await
 import scala.concurrent.duration._
@@ -51,7 +47,8 @@ class NoopTrustManager extends X509TrustManager {
   }
 }
 
-class PubSubApiSpec extends AnyFlatSpec with BeforeAndAfterAll with ScalaFutures with Matchers with LogCapturing {
+class PubSubApiSpec extends AnyFlatSpec with BeforeAndAfterAll with ScalaFutures with Matchers with LogCapturing
+    with PubSubMockSpec {
 
   implicit val system: ActorSystem = ActorSystem(
     "PubSubApiSpec",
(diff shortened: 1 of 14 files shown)


google-cloud-pub-sub/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/impl/PubSubApiSpec.scala
mdedetrich, line 1: Since `NoCredentials` no longer supplies a token, all of the tests in this suite needed to be fixed by removing the authorization header checks.

`AuthorizationHeaderSpec` was created to make sure that google-pub-sub is still calling google-commons `addAuth` to add the authorization header to prevent any possible future regressions.

google-common/src/test/scala/org/apache/pekko/stream/connectors/google/PaginatedRequestSpec.scala
mdedetrich, line 1: Note that the authorization header is still being tested in `GoogleHttpSpec`

google-common/src/test/scala/org/apache/pekko/stream/connectors/google/ResumableUploadSpec.scala
mdedetrich, line 1: Note that the authorization header is still being tested in `GoogleHttpSpec`

google-fcm/src/test/scala/org/apache/pekko/stream/connectors/google/firebase/fcm/v1/impl/FcmSenderSpec.scala
mdedetrich, line 1: I didn't bother adding a test here to make sure that the authorization header is being added for non `NoCredentials` credentials provider but I ca do it on request.

Its also unclear if these are meant to be proper tests or tests asserting that the docs are written clearly???

pjfanning:
Is it ok to call this a v1.2.0 change? I have no problem with making such a release the next one for this module.

mdedetrich:
Im putting this into draft mode as there is actually a legitimate use for an auth method that is just project-id+token, its just that calling this authentication method `none` is quite a misnomer and such an auth method should not be used in tests.

I will update the PR to add another auth method (such as `token`) which has the same behaviour as the current `none`. With this change it definitely should land for 1.2.x

mdedetrich:
@pjfanning  The PR is now ready to review, essentially the behaviour of previous `none` (i.e. `NoCredentials`) has been changed to `access-token` (i.e. `AccessTokenCredentials`) and `none` now doesn't provide/expect a token (which has been unchanged in this PR).

pjfanning (approved):
lgtm

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/AccessTokenCredentials.scala
```diff
@@ -0,0 +1,50 @@
...
+import com.google.auth.{ Credentials => GoogleCredentials }
+import com.typesafe.config.Config
+
+import java.net.URI
+import java.util
+import scala.concurrent.{ ExecutionContext, Future }
+
+@InternalApi
+private[auth] object AccessTokenCredentials {
+  def apply(c: Config): AccessTokenCredentials = AccessTokenCredentials(c.getString("project-id"), c.getString("token"))
```
He-Pin, line 29: we may need a java api with create method

He-Pin (approved):

samueleresca (approved):

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/auth/AccessTokenCredentials.scala
mdedetrich: This is private, its not meant to be created by java users and its only being used internally within Scala API (no other auth data classes have a java create method either as they are all private)

PR #1026, opened 2025-03-30
Properly propogate GoogleSettings in addStandardQuery

mdedetrich:
When working on adding tests that run on [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) I programmatically configure `GoogleSettings` rather than having it resolve via typesafe config since I am using testcontaines with dynamic values for the host (this is follow up from https://github.com/apache/pekko-connectors/pull/1020 and https://github.com/apache/pekko-connectors/pull/1022).

I stumbled upon tests failing due to them resolving `GoogleSettings` via [`application.conf`](https://github.com/apache/pekko-connectors/blob/1fe3474d654bf48b0e81011d878e541a32409c93/google-cloud-storage/src/test/resources/application.conf) even though I was explicitly trying to bypass this type of typesafe config loading i.e.

```scala
lazy val googleSettings =
  GoogleSettings.create(
    "test-project",
    NoCredentials("", ""),
    RequestSettings.create(
      Optional.empty(),
      Optional.empty(),
      prettyPrint = false,
      15728640, // 15 MiB
      RetrySettings.create(
        6,
        Duration.of(1, ChronoUnit.SECONDS),
        Duration.of(1, ChronoUnit.MINUTES),
        0.2d
      ),
      Optional.empty()
    )
  )
```

In regards to settings/attributes, typically the way that the core code works is that we resolve settings (in this case `GoogleSettings`) and then we pass these settings along using Scala's implicit parameter passing. Normally Scala's implicits will create a compiler error if an implicit parameter is not found, but in this specific case there is an implicit default fallback defined at https://github.com/apache/pekko-connectors/blob/e0202ad934f8a218bcded64835fb2ab99ead6e6f/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/GoogleSettings.scala#L78 which means that if you forget to pass along the implicit it will resolve

```scala
implicit def settings(implicit system: ClassicActorSystemProvider): GoogleSettings = apply(system)
```

Since there always happens to be a `system` in scope.

In this specific case there is a [addStandardQuery](https://github.com/apache/pekko-connectors/blob/fc4417af82c7c22439cf6f7e5a1e87298d0e2107/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/http/GoogleHttp.scala#L158) function which does specify it needs `GoogleSettings` however neither of the callers of this function (`singleRequest`/`singleRawRequest`) actually pass the implicit `GoogleSettings` along hence causing it to fallback the default provider that is resolved via typesafe config (via the `system`)

The fix for this is simple, just need to add `googleSettings: GoogleSettings` to the implicit parameter list of these functions.

### DIFF
--- a/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/http/GoogleHttp.scala
+++ b/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/http/GoogleHttp.scala
@@ -52,7 +52,8 @@ private[connectors] final class GoogleHttp private (val http: HttpExt) extends A
   /**
    * Sends a single [[HttpRequest]] and returns the raw [[HttpResponse]].
    */
-  def singleRawRequest(request: HttpRequest)(implicit settings: RequestSettings): Future[HttpResponse] = {
+  def singleRawRequest(request: HttpRequest)(implicit settings: RequestSettings, googleSettings: GoogleSettings)
+      : Future[HttpResponse] = {
     val requestWithStandardParams = addStandardQuery(request)
     settings.forwardProxy.fold(http.singleRequest(requestWithStandardParams)) { proxy =>
       http.singleRequest(requestWithStandardParams, proxy.connectionContext, proxy.poolSettings)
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

but should we make the next release 1.2.0 due to the signature change?

mdedetrich:
> lgtm
> 
> 
> 
> but should we make the next release 1.2.0 due to the signature change?

This is all internal functions, none of it is public

PR #1032, opened 2025-04-10
Bucket name should be pekko-connectors

mdedetrich:
This as as per instructions at https://github.com/apache/pekko-connectors/blob/7b5878e46ee76b39619b92b1175e1af40af0eccf/google-cloud-storage/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/storage/impl/GoogleGCStorageStreamIntegrationSpec.scala#L35


pjfanning (approved):
lgtm

PR #1033, opened 2025-04-10
add 1.1.x branch

pjfanning:
relates to discussion about 1.2.0 release in #1021

### DIFF
--- a/.github/workflows/nightly-builds.yaml
+++ b/.github/workflows/nightly-builds.yaml
@@ -28,22 +28,6 @@ jobs:
   test-compile:
     name: Compile
     runs-on: ubuntu-22.04
-    # Licensed to the Apache Software Foundation (ASF) under one
-# or more contributor license agreements.  See the NOTICE file
-# distributed with this work for additional information
-# regarding copyright ownership.  The ASF licenses this file
-# to you under the Apache License, Version 2.0 (the
-# "License"); you may not use this file except in compliance
-# with the License.  You may obtain a copy of the License at
-#
-#   http://www.apache.org/licenses/LICENSE-2.0
-#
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly-builds.yaml
```diff
@@ -28,22 +28,6 @@ jobs:
   test-compile:
     name: Compile
     runs-on: ubuntu-22.04
-    # Licensed to the Apache Software Foundation (ASF) under one
```
pjfanning, line 31: license header pasted here in error - it appears on top of file anyway

mdedetrich:
I have a few more bug fixes for 1.1.x  so shall we merge this after those bug fixes land?

mdedetrich:
> > I have a few more bug fixes for 1.1.x so shall we merge this after those bug fixes land?
> 
> maybe - all the same, I'd be as happy to just get everything into a 1.2.0 release and not worry so much about a 1.1.1 release

So regarding https://github.com/apache/pekko-connectors/pull/1021, Im actually going to put it on hold/into draft mode because there is actually a use case for heaving an auth method with just project_id + token its just that it should be called something else (i.e. token auth method) since I am using the token auth method against GCS proper right now

In general I was actually contemplating about whether we should start splitting out the various components of connectors into their own repos. Doing a release for 1.1.1 makes sense since bug fixes have been sprinkled around the various connectors but 1.2.x would be a big release for only a few connectors, wdyt?

pjfanning:
I'm not sure that a big refactor of the repo and splitting it into sub-repos is a 1.x.y task - seems more like a major release task.

mdedetrich (approved):
ill approve this for now, its too soon to split out the mono repo for now

Issue #1009, opened 2025-03-21
MQTT 5.0 support in pekko-connectors-mqtt


sndnv:
This is more of a question and a pre-PR discussion rather than a bug or a feature request.

---

Currently, the MQTT connector supports only MQTT v3 using the [`org.eclipse.paho.client.mqttv3`](https://github.com/apache/pekko-connectors/blob/main/project/Dependencies.scala#L419) dependency.

After a quick look at the Paho clients, I don't get the impression that the v5 client is able to work with a v3 broker (or a v3 client with a v5 broker) so if support for v5 is to be added to Pekko, I assume both v3 and v5 need to be supported for the foreseeable future.

My question is, what would be the preferable way of approaching this? Ideally, both dependencies/clients would coexist in `pekko-connectors-mqtt` and some config option would decide which one to go with but it could be that there are some dependency conflicts or that the APIs are so vastly different that it doesn't make sense to put them together (I didn't check that myself yet). So, what would be an acceptable alternative? Have `pekko-connectors-mqtt` as the v3 client and something like `pekko-connectors-mqttv5` as the v5 one?

pjfanning:
I suspect a new connector is the way to go.

sndnv:
Resolved by #1035

PR #1035, opened 2025-04-11
Add MQTT v5 support

sndnv:
As briefly discussed in #1009 this PR adds a new module called `mqttv5` with support for MQTT v5.

As it is right now, it does violate the DRY principle a bit by copying everything (unmodified) from:
- `mqtt.javadsl`
- `mqtt.scaladsl`
- `mqtt.model`

So my question is, what would be the best way to move forward? I see the following options:
- keep as is - as a drop-in replacement of the v3 version, it works; the effort to switch to `mqttv5` is minimal (the package name is different and a couple of the `withXXX` methods are gone but that's all)
- make the `mqttv5` module support both v3 and v5 and deprecate the current `mqtt` module

Making the current `mqtt` module support both v3 and v5 is not going to work without breaking source compatibility for somebody.

### DIFF
--- a/mqttv5/src/main/scala/org/apache/pekko/stream/connectors/mqttv5/impl/MqttFlowStage.scala
+++ b/mqttv5/src/main/scala/org/apache/pekko/stream/connectors/mqttv5/impl/MqttFlowStage.scala
@@ -1,0 +1,516 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+/*
+ * Copyright (C) since 2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package org.apache.pekko.stream.connectors.mqttv5.impl
+
+import java.util.concurrent.Semaphore
+import java.util.concurrent.atomic.AtomicInteger
+
+import scala.collection.mutable
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+import scala.util.control.NonFatal
+
+import org.apache.pekko.Done
+import org.apache.pekko.annotation.InternalApi
+import org.apache.pekko.stream.Shape
+import org.apache.pekko.stream._
+import org.apache.pekko.stream.connectors.mqttv5.AuthSettings
+import org.apache.pekko.stream.connectors.mqttv5.MqttConnectionSettings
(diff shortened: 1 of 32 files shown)


sndnv:
🤔 Do I need to add the new module [here](https://github.com/apache/pekko-connectors/blob/main/build.sbt#L479) to skip the `mima` checks? The compatibility build/step is failing because of it.

pjfanning:
> 🤔 Do I need to add the new module [here](https://github.com/apache/pekko-connectors/blob/main/build.sbt#L479) to skip the `mima` checks? The compatibility build/step is failing because of it.

That should fix the mima check.

mdedetrich:
I am personally more inline with the lets just copy for now until we know that its stable and working, and then we can always do some code reuse between the modules seeing as most of the cruft should be internal API (and if its not then by definition it should be quite segregated).

pjfanning:
@sndnv thank you for the large contribution. I haven't had a full look yet but it looks like it should be mergeable pretty much as is. Since it is a large change, would you be able to sign a CLA? See https://www.apache.org/licenses/contributor-agreements.html

sndnv:
> @sndnv thank you for the large contribution. I haven't had a full look yet but it looks like it should be mergeable pretty much as is. Since it is a large change, would you be able to sign a CLA? See https://www.apache.org/licenses/contributor-agreements.html

Signed; not sure what happens next but the email is sent.

---

Since the agreement is to keep the module as-is for now, I'll add a few comments to the diff to make the review easier. Most of that code is a straight copy from `mqtt`, including the tests.

mqttv5/src/main/scala/org/apache/pekko/stream/connectors/mqttv5/impl/MqttFlowStage.scala
```diff
@@ -0,0 +1,516 @@
...
+    pahoMsg.setQos(msg.qos.getOrElse(defaultQoS).value)
+    pahoMsg.setRetained(msg.retained)
+
+    mqttClient.publish(
+      msg.topic,
+      pahoMsg,
+      msg,
+      /* callback */ new MqttActionListener {
+        def onSuccess(token: IMqttToken): Unit = {
+          token.getReasonCodes.toList.filter(_ >= MqttReturnCode.RETURN_CODE_UNSPECIFIED_ERROR).distinct match {
```
sndnv, line 412: [info] As with the subscription above, the return codes are checked and the stage is failed if something went wrong.

pjfanning:
> @sndnv thank you for the large contribution. I haven't had a full look yet but it looks like it should be mergeable pretty much as is. Since it is a large change, would you be able to sign a CLA? See https://www.apache.org/licenses/contributor-agreements.html

The iCLA has been registered by the ASF Secretary. Thanks.

pjfanning (approved):
lgtm - thanks

PR #1045, opened 2025-04-23
Fix metadata not working with GCStorage

mdedetrich:
Resolves https://github.com/apache/pekko-connectors/issues/656 .

Underlying issue is that metadata needs to be nested inside of its own "metadata" object, see https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request-body

With this test, running `sbt "google-cloud-storage/Test/runMain org.scalatest.tools.Runner -o -s org.apache.pekko.stream.connectors.googlecloud.storage.impl.GoogleGCStorageStreamIntegrationSpec"` against an actual  google cloud account works.


pjfanning (approved):
lgtm

PR #1060, opened 2025-05-05
Make GCS resumableUpload work with empty byte payload

mdedetrich:
Resolves: https://github.com/apache/pekko-connectors/issues/1054

This PR solves resumable upload for gcs when uploading an empty payload (i.e. zero bytes).  Comments have been added explaining how the solution works, but in short a special sentinel value gets inserted in the `onComplete` of `statefulMap` where the chunks get calculated on the specific case that the state is empty (which means nothing is being processed, hence empty payload) and this sentinel value gets used/detected later when calculating the Content-Range header.

A test has been added to confirm that this works, without this change the newly added test fails. I also rerun the entire test suite against a real GCS account to make sure there are no additional regressions.

@He-Pin Would be good if you check this PR as you originally wrote some of the code being touched

### DIFF
--- a/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
+++ b/google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
@@ -17,11 +17,12 @@ import org.apache.pekko
 import pekko.NotUsed
 import pekko.annotation.InternalApi
 import pekko.dispatch.ExecutionContexts
+import pekko.http.scaladsl.model.ContentRange.Unsatisfiable
 import pekko.http.scaladsl.model.HttpMethods.{ POST, PUT }
 import pekko.http.scaladsl.model.StatusCodes.{ Created, OK, PermanentRedirect }
 import pekko.http.scaladsl.model._
 import pekko.http.scaladsl.model.headers.ByteRange.Slice
-import pekko.http.scaladsl.model.headers.{ `Content-Range`, Location, Range, RawHeader }
+import pekko.http.scaladsl.model.headers.{ `Content-Range`, Location, Range, RangeUnits, RawHeader }
 import pekko.http.scaladsl.unmarshalling.{ FromResponseUnmarshaller, Unmarshal, Unmarshaller }
 import pekko.stream.Materializer
 import pekko.stream.connectors.google.http.GoogleHttp
@@ -39,6 +40,10 @@ private[connectors] object ResumableUpload {
   final case class InvalidResponseException(override val info: ErrorInfo) extends ExceptionWithErrorInfo(info)
   final case class UploadFailedException() extends Exception
   private final case class Chunk(bytes: ByteString, position: Long)
+
+  private object Chunk {
+    final val zero = Chunk(ByteString.empty, 0)
+  }
 
   /**
    * Initializes and runs a resumable upload to a media endpoint.
@@ -65,7 +70,7 @@ private[connectors] object ResumableUpload {
           .statefulMap(() => 0L)((cumulativeLength, bytes) =>
               (cumulativeLength + bytes.length, Chunk(bytes, cumulativeLength)),
             _ => None)
-          .via(AnnotateLast[Chunk])
+          .via(AnnotateLast[Chunk](Chunk.zero))
           .map(chunk => Future.successful(Right(chunk)))
 
         val upload = Flow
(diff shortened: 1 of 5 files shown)


google-common/src/test/scala/org/apache/pekko/stream/connectors/google/util/AnnotateLastSpec.scala
```diff
@@ -33,22 +33,23 @@ class AnnotateLastSpec
...
-      val probe = Source.single(1).via(AnnotateLast[Int]).runWith(TestSink.probe)
+      val probe = Source.single(1).via(AnnotateLast[Int](0)).runWith(TestSink.probe)
       probe.requestNext(Last(1))
       probe.expectComplete()
     }
 
-    "do nothing when stream is empty" in {
-      val probe = Source.empty[Nothing].via(AnnotateLast[Nothing]).runWith(TestSink.probe)
-      probe.expectSubscriptionAndComplete()
+    "return zero value when stream is empty" in {
```
mdedetrich, line 49: Note that while this PR does change the behaviour of `AnnotateLast`, this doesn't appear to be used outside of the context of `ResumableUpload` in gcs. If requested, it is possible to provide 2 versions of `AnnotateLast.apply`, one that emits the sentinel value in case of the empty stream and another that doesn't emit anything on empty stream.

mdedetrich:
I have just updated the tests so that we also make sure that the returning `StorageObject` reports that the uploaded file is 0 bytes in size and has a static md5hash of those empty bytes

mdedetrich:
@pjfanning @He-Pin @raboof @Roiocam @nvollmar @gmethvin Sorry for the nag but can one of you review this PR? I want to get it tested by a client of mine who experienced this bug in production to confirm that it works and to do so a snapshot needs to be published.

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/util/AnnotateLast.scala
```diff
@@ -45,14 +45,33 @@ private[google] final case class Last[+T](value: T) extends MaybeLast[T] {
...
+        _.map(elem => Some(Last(elem)))
+      ).collect {
+        case Some(elem) => elem
+      }
+
+  def apply[T](zero: T): Flow[T, MaybeLast[T], NotUsed] =
+    Flow[T]
+      .statefulMap(() => Option.empty[T])(
+        accumulateState,
+        {
```
He-Pin, line 67: Maybe the old style is simpler than the anymouse one?

He-Pin (approved):
overall lgtm

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/util/AnnotateLast.scala
mdedetrich: I prefer this one as there is less boilerplate

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
```diff
@@ -65,7 +66,7 @@ private[connectors] object ResumableUpload {
           .statefulMap(() => 0L)((cumulativeLength, bytes) =>
               (cumulativeLength + bytes.length, Chunk(bytes, cumulativeLength)),
             _ => None)
-          .via(AnnotateLast[Chunk])
+          .via(AnnotateLast[Chunk](Chunk(ByteString.empty, 0)))
```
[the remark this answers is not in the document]
mdedetrich: Done and pushed

mdedetrich:
Thanks for the review, I extracted the zero chunk as a constant as asked. If tests pass I will merge and then test this.

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/util/AnnotateLast.scala
He-Pin: that's ture

He-Pin (approved):

google-common/src/main/scala/org/apache/pekko/stream/connectors/google/ResumableUpload.scala
```diff
@@ -40,6 +41,8 @@ private[connectors] object ResumableUpload {
   final case class UploadFailedException() extends Exception
   private final case class Chunk(bytes: ByteString, position: Long)
 
+  private val zeroChunk = Chunk(ByteString.empty, 0)
```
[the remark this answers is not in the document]
mdedetrich: Done and pushed

PR #1075, opened 2025-05-19
CI: use sha for actions instead of version

pjfanning:
see https://github.com/apache/pekko-management/pull/429


raboof (approved):
(haven't checked that the hashes match the versions :) )

PR #1089, opened 2025-06-15
move away from sbt IntegrationTest config

pjfanning:
sbt has deprecated (for removal) the IntegrationTest config


danischroeter (approved):
lgtm

PR #1091, opened 2025-06-15
junit-interface 0.13.3

pjfanning:
has moved to sbt org


raboof:
scala-steward should already know about this migration (https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/resources/artifact-migrations.v2.conf#L705) - I wonder why it wasn't picked up?

pjfanning:
@raboof is it ok to merge this? ScalaSteward is good but it does miss some stuff. I had to create #1088 too - just something else that ScalaSteward missed.

raboof:
I'd rather fix it in scala-steward, but realistically I won't have time for that any time soon, so I've created an issue upstream for now to not forget that we encountered this here. with that I can think we can merge this.

PR #1103, opened 2025-06-29
Manage Pekko upgrades explicitly

pjfanning:
Failures in CI because we can't have Scala Steward upgrade CI actions (due to how ASF configure our repos)
  https://github.com/apache/pekko-connectors/actions/runs/15950739121

Easier at the moment to manage pekko upgrades using manual PRs.
The Scala Steward failures are blocking us from getting PRs for some other libs that are already out of date.

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -42,6 +42,8 @@ updates.pin  = [
 ]
 
 updates.ignore = [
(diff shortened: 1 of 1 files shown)


pjfanning:
@kerr would you have time to review this?

PR #1189, opened 2025-09-12
test with pekko 1.2.x

pjfanning:
* we've seen at least 1 bin compat issue with Pekko 1.2.0
* we don't really need to test pekko 1.0.x nightly
* update action versions
* change the nightly tests to test the 1.2.x branch as we need a 1.2.0 release but little point testing main branch yet because we are making breaking changes


He-Pin (approved):
lgtm

PR #1198, opened 2025-09-14
use scala option converter

pjfanning:
there are a few collection converter changes too but there are lots more that I intend to do in a follow up


He-Pin (approved):
lgtm

PR #1205, opened 2025-09-19
Avoid silently swallowing the error when SSE stream fails/retries

rrwright:
# Problem

The current pekko-http SSE (Server-Sent Events) client implementation fails its stream every time a message arrives over the size limits. When a Pekko SSE client connects to a Pekko SSE server who serves a message larger than the client's size limit, the pekko-http client fails the stream, then the SSE Connector tries to reconnect and resume at the same (oversized) message which fails again. The SSE Connector client falls into a silent infinite connection retry loop. The oversized message error is swallowed by the SSE Connector, so the user never knows what is happening. The observed behavior appears as if the stream is still active and just gets stuck at the message that preceded the oversized message. Meanwhile the SSE connector is infinitely retrying… silently.

# Solution
This simple PR just avoids silently swallowing the failure message, and instead logs an error. The existing retry logic is left untouched (but really should be designed better for useful error handling).

Another PR to the pekko-http library has been submitted to add additional options beyond just failing stream when either an SSE line or event exceeds the size limit. https://github.com/apache/pekko-http/pull/744  Some of the new options in that PR can avoid the stream failing in the first place (example: `log-and-skip`). So this PR is meant to stay as small as possible, with related key new functionality added to the core pekko-http library.

### DIFF
--- a/sse/src/main/scala/org/apache/pekko/stream/connectors/sse/scaladsl/EventSource.scala
+++ b/sse/src/main/scala/org/apache/pekko/stream/connectors/sse/scaladsl/EventSource.scala
@@ -17,6 +17,7 @@ package scaladsl
 import org.apache.pekko
 import pekko.NotUsed
 import pekko.actor.{ ActorSystem, ClassicActorSystemProvider }
+import pekko.event.Logging
 import pekko.http.scaladsl.client.RequestBuilding.Get
 import pekko.http.scaladsl.coding.Coders
 import pekko.http.scaladsl.model.MediaTypes.`text/event-stream`
@@ -99,6 +100,7 @@ object EventSource {
     import EventStreamUnmarshalling.fromEventsStream
     implicit val actorSystem: ActorSystem = system.classicSystem
     import actorSystem.dispatcher
+    val log = Logging(actorSystem, "EventSource")
 
     val continuousEvents = {
       def getEventSource(lastEventId: Option[String]) = {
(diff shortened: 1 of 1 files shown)


sse/src/main/scala/org/apache/pekko/stream/connectors/sse/scaladsl/EventSource.scala
```diff
@@ -99,6 +100,7 @@ object EventSource {
     import EventStreamUnmarshalling.fromEventsStream
     implicit val actorSystem: ActorSystem = system.classicSystem
     import actorSystem.dispatcher
+    val log = Logging(actorSystem, getClass)
```
pjfanning, line 103: can this be private?

rrwright: It's in the body of the `apply` method, so it's already is inaccessible even without the `private` keyword.

sse/src/main/scala/org/apache/pekko/stream/connectors/sse/scaladsl/EventSource.scala
```diff
@@ -111,7 +113,12 @@ object EventSource {
           .flatMap(Unmarshal(_).to[EventSource])
           .fallbackTo(Future.successful(noEvents))
       }
-      def recover(eventSource: EventSource) = eventSource.recoverWithRetries(1, { case _ => noEvents })
+      def recover(eventSource: EventSource) = eventSource.recoverWithRetries(1,
+        {
+          case e =>
+            log.warning(e, "SSE Connector is retrying failed stream for: {} ", uri)
```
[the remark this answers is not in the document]
raboof: > The best code design would probably include some user-level choice about whether to silently retry forever or be notified. I tried to stay away from editing the code logic… otherwise I'd want to rewrite most of it. :-)

:)

Thanks for your thoughtful changes!

raboof (approved):

PR #1226, opened 2025-09-28
pin activemq (used only in JMS tests)

pjfanning:
see #1194


raboof (approved):
Makes sense to me. We should explain all this in the docs but that can be a separate PR

PR #1246, opened 2025-10-16
AWS S3: Handle Illegal Headers in Copy Part

### LINKED ISSUE #1245: MultipartCopy Fails with the presence of Storage Class Header
An extension of: 
https://github.com/apache/pekko-connectors/pull/81

When `x-amz-storage-class` is manually added to the headers of a `multipartCopy` request the resulting call fails with a 

```
S3Exception: The specified header is not valid in this context
```

I believe the fix is the same as what was in the original PR, however it was not done for the CopyPart

Changing the code to follow the pattern of the linked PR resolves the issue - however it's worth noting that this does come at the expense of other legitimate fields like `x-amz-request-payer`

[See AWS docs for valid headers](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html#API_UploadPartCopy_RequestSyntax)

It seems that we would need to filter out the headers that are not in this list, rather than only include the SSE headers.

Happy to contribute - although I'm not exactly sure what would be the pattern to employ to only allow for a subset of headers

Luka-J9:
Currently this is a minimal change to demonstrate the issue so that the strategy can be discussed before addressed.

fixes #1245

### DIFF
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Request.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Request.scala
@@ -13,8 +13,10 @@
 
 package org.apache.pekko.stream.connectors.s3.impl
 
-import org.apache.pekko.annotation.InternalApi
-import org.apache.pekko.util.OptionVal
+import org.apache.pekko
+
+import pekko.annotation.InternalApi
+import pekko.util.OptionVal
 
 /**
  * Internal Api
--- a/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/settings.scala
+++ b/s3/src/main/scala/org/apache/pekko/stream/connectors/s3/settings.scala
@@ -15,7 +15,6 @@ package org.apache.pekko.stream.connectors.s3
 
 import com.typesafe.config.Config
 import org.apache.pekko
-import org.apache.pekko.stream.connectors.s3.impl.S3Request
 import org.slf4j.LoggerFactory
 import software.amazon.awssdk.auth.credentials._
 import software.amazon.awssdk.regions.Region
@@ -36,6 +35,7 @@ import pekko.actor.{ ActorSystem, ClassicActorSystemProvider }
 import pekko.http.scaladsl.model.Uri
 import pekko.stream.connectors.s3.AccessStyle.{ PathAccessStyle, VirtualHostAccessStyle }
 import pekko.util.OptionVal
+import pekko.stream.connectors.s3.impl.S3Request
 
 final class Proxy private (
     val host: String,
--- a/s3/src/test/scala/org/apache/pekko/stream/connectors/s3/S3HeadersSpec.scala
+++ b/s3/src/test/scala/org/apache/pekko/stream/connectors/s3/S3HeadersSpec.scala
@@ -18,9 +18,11 @@
 package org.apache.pekko.stream.connectors.s3
 
 import com.typesafe.config.ConfigFactory
-import org.apache.pekko.stream.connectors.s3.impl._
+import org.apache.pekko
 import org.scalatest.flatspec.AnyFlatSpecLike
 import org.scalatest.matchers.should.Matchers
+
+import pekko.stream.connectors.s3.impl._
 
 class S3HeadersSpec extends AnyFlatSpecLike with Matchers {
   it should "filter headers based on what's allowed" in {


pjfanning:
I have no strong opinion about being proactive about headers. It seems to introduce its own risks. If we do introduce a new config, avoid any name that uses colour based naming, eg white or black.

Luka-J9:
@mdedetrich / @pjfanning - Do I need to worry about source/binary compat? Not sure I can add the configuration handling without breaking compatibility on some level

Additionally I assume I should be handling the headers the same way for all header S3Request types and not just CopyPart?

pjfanning:
> @mdedetrich / @pjfanning - Do I need to worry about source/binary compat? Not sure I can add the configuration handling without breaking compatibility on some level
> 
> Additionally I assume I should be handling the headers the same way for all header S3Request types and not just CopyPart?

@Luka-J9 it's ok to break binary compatibility on main branch as it is targeted at a 2.0.0 release. Just add a mima-filters file. You'll find examples in the src code already (`.excludes` file extension).
We can decide based on whether it is safe to backport to a 1.3.0 release after that. We would accept some level of change for that but it would probably need careful thought as to what is acceptable.

mdedetrich:
> @mdedetrich / @pjfanning - Do I need to worry about source/binary compat? Not sure I can add the configuration handling without breaking compatibility on some level
> 
> Additionally I assume I should be handling the headers the same way for all header S3Request types and not just CopyPart?

So I am not sure how this change would be binary or even source breaking, it should be just internal behaviour without needing to change and function/method signatures. We already have a mima check that will verify if there is a binary breaking change.

Regarding adding the configuration setting, this is fine as long as its released in a version with a bumped minor version.

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Stream.scala
```diff
@@ -832,7 +829,6 @@ import scala.util.{ Failure, Success, Try }
       bucket: String,
       method: HttpMethod,
       httpRequest: (HttpMethod, S3Settings) => HttpRequest,
-      headers: Seq[HttpHeader],
```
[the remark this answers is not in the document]
Luka-J9: So it is actually important to remove this as not doing so makes it difficult or really messy to maintain backwards compat.

mdedetrich:
> > > > @Luka-J9 I just ran the CI and some S3 tests failed, would you be able to look into it?
> > > 
> > > 
> > > Looks like one of my tests is not compatible with scala 3 - it uses reflection to get a list of members of a sealed trait - I'll see if I can work around it
> > 
> > 
> > There are open source libraries that target both Scala 2 and Scala 3 which can do this (i.e. https://github.com/zio/izumi-reflect), albeit this can be considered a bit excessive.
> > If you do decide to use this library make sure to add the dependency only to `test` scope as we don't need as part of main s3.
> 
> Opted to use exhaustivity checking - it should be good for another CI run now 🤞

Code is not formatted, format with scalafmt

Luka-J9:
> > > > > @Luka-J9 I just ran the CI and some S3 tests failed, would you be able to look into it?
> > > > 
> > > > 
> > > > Looks like one of my tests is not compatible with scala 3 - it uses reflection to get a list of members of a sealed trait - I'll see if I can work around it
> > > 
> > > 
> > > There are open source libraries that target both Scala 2 and Scala 3 which can do this (i.e. https://github.com/zio/izumi-reflect), albeit this can be considered a bit excessive.
> > > If you do decide to use this library make sure to add the dependency only to `test` scope as we don't need as part of main s3.
> > 
> > 
> > Opted to use exhaustivity checking - it should be good for another CI run now 🤞
> 
> Code is not formatted, format with scalafmt

Missed a newline when updating the reflection test 🤦 
Sorted imports while I was at it

mdedetrich:
Awesome thanks! Ill have a look at this on the weekend.

s3/src/main/resources/reference.conf
```diff
@@ -125,4 +125,20 @@ pekko.connectors.s3 {
 
   # Add signature headers to requests when aws.credentials.provider is anon
   sign-anonymous-requests = true
+
+  additional-allowed-headers {
```
mdedetrich: Can we add a comment here so readers have context?

s3/src/main/resources/reference.conf
```diff
@@ -125,4 +125,20 @@ pekko.connectors.s3 {
...
+    GetObject = []
```
mdedetrich: I might be missing something, but are these the right defaults? It feels like by default there should be some entries here (since you are testing them) and if someone doesn't want these additional allowed headers then they can override it in their own `.conf` with an empty list.

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Request.scala
```diff
@@ -18,69 +18,400 @@ import org.apache.pekko.annotation.InternalApi
...
  */
-@InternalApi private[s3] case object PutObject extends S3Request
+@InternalApi private[s3] case object HeadObject extends S3Request {
+
+  /**
+   * See [[https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_RequestSyntax HeadObject Request Syntax]]
+   * @return all valid headers for HeadObject request type
+   */
+  override def allowedHeaders: Set[String] = Set(
+    "Host",
```
Luka-J9: @mdedetrich - here is an example of the "base" headers derived from the aws docs. Do you want me to pull this into the .conf file?

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/impl/S3Request.scala
```diff
@@ -23,64 +23,130 @@ import org.apache.pekko.annotation.InternalApi
...
+      case "CopyPart"                => Some(CopyPart)
+      case "DeleteObject"            => Some(DeleteObject)
+      case "ListBucket"              => Some(ListBucket)
+      case "MakeBucket"              => Some(MakeBucket)
+      case "DeleteBucket"            => Some(DeleteBucket)
+      case "CheckBucket"             => Some(CheckBucket)
+      case "PutBucketVersioning"     => Some(PutBucketVersioning)
+      case "GetBucketVersioning"     => Some(GetBucketVersioning)
+      case _                         => None
+    }
```
He-Pin: If we use OptionVal to avoid allocation

He-Pin (approved):
Lgtm over all

He-Pin:
Thank you for this amazing contribution

He-Pin:
@Luka-J9 Seems there are some typos

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/settings.scala
```diff
@@ -420,6 +427,17 @@ final class S3Settings private (
   def withSignAnonymousRequests(value: Boolean): S3Settings =
     if (signAnonymousRequests == value) this else copy(signAnonymousRequests = value)
 
+  private[s3] val concreateAllowedHeaders: Map[S3Request, Set[String]] = {
+    allowedHeaders.foldLeft(Map.empty[S3Request, Set[String]]) {
+      case (acc, (header, value)) =>
+        S3Request.fromString(header) match {
+          case OptionVal.Some(header) => acc + (header -> value)
+          case OptionVal.None         => acc
+          case other                  => throw new MatchError(other)
```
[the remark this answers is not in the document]
Luka-J9: This isn't valid - additionally other parts of code does it this way

s3/src/main/scala/org/apache/pekko/stream/connectors/s3/settings.scala
```diff
@@ -13,25 +13,30 @@
...
-import java.util.concurrent.TimeUnit
-import java.util.{ Objects, Optional }
-
-import org.apache.pekko
-import pekko.actor.{ ActorSystem, ClassicActorSystemProvider }
-import pekko.http.scaladsl.model.Uri
-import pekko.stream.connectors.s3.AccessStyle.{ PathAccessStyle, VirtualHostAccessStyle }
 import com.typesafe.config.Config
+import org.apache.pekko
+import org.apache.pekko.stream.connectors.s3.impl.S3Request
```
pjfanning: If you import org.apache.pekko, then can omit the org.apache bit in later pekko imports. This is preferred.

mdedetrich:
Ill go ahead and merge/cherry pick PR

PR #1400, opened 2026-02-03
update java format

pjfanning:
* relates to issues seen in #1379 
* aim is to merge this and then add an entry to ignore the commit in git blame
* then 1379 should be mergeable without needing its own java reformatting

This runs ok on my laptop but I see it fails on CI.

I suspect at this stage that we would be better off turning off the Java formatting check in CI for this repo for the time being.

### DIFF
--- a/amqp/src/test/java/docs/javadsl/AmqpDocsTest.java
+++ b/amqp/src/test/java/docs/javadsl/AmqpDocsTest.java
@@ -13,6 +13,14 @@
 
 package docs.javadsl;
 
+import static org.junit.Assert.assertEquals;
+
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import org.apache.pekko.Done;
 import org.apache.pekko.NotUsed;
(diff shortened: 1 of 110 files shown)


raboof:
> This runs ok on my laptop but I see it fails on CI.

Ugh. Also generally I'm not sure we _want_ to organize the imports, as these are sometimes grouped for inclusion in the docs. It seems [upstream](https://github.com/google/google-java-format) does have a `--skip-sort-imports` option, but that's not exposed through sbt-java-formatter yet AFAICS. Filed https://github.com/sbt/sbt-java-formatter/issues/251 . 

> I suspect at this stage that we would be better off turning off the Java formatting check in CI for this repo for the time being.

I'd be OK with that.

mkurz:
> It seems [upstream](https://github.com/google/google-java-format) does have a `--skip-sort-imports` option,

You can use that now with latest sbt-java-formatter RC: https://github.com/sbt/sbt-java-formatter/releases/tag/v0.12.0-RC1

Please read the release notes and give it a try.
Example how to set the env variable on CI can be found in this diff: https://github.com/sbt/sbt/pull/9098

Please test and let me know what you think!
Thanks!

mkurz:
- See #1588

mkurz:
can be closed i guess now that #1588 was merged

PR #1450, opened 2026-02-24
change nightly job to run doc publish as an independent subjob

pjfanning:
#1441 did not fix the nightly build
That PR shows that the doc build when it runs independently so it now looks like running it after a jar publish job seems to affect the doc build.


raboof (approved):
weird but OK

PR #1480, opened 2026-03-05
MQTT v5: expose MQTT 5 user properties on MqttMessage

### LINKED ISSUE #1371: MQTT v5 connector does not expose MQTT 5 message properties
We would like to use the Apache Pekko MQTT v5 connector in production, but we are currently blocked from doing so because MQTT 5 message properties are not exposed.

While the underlying Paho MQTT v5 client provides access to MQTT 5 message properties (e.g. User Properties), these properties are currently dropped by the Pekko MQTT v5 connector.

In the connector’s message model
`org.apache.pekko.stream.connectors.mqttv5.model.MqttMessage`,
only the topic and payload are exposed. MQTT 5 properties, such as User Properties, are not part of the model and therefore cannot be accessed by stream consumers.

This prevents us from accessing required metadata (e.g. business identifiers and tracing information) that we transport via MQTT 5 User Properties and currently blocks us from using the MQTT v5 connector beyond a proof of concept.

The underlying Paho `org.eclipse.paho.mqttv5.common.MqttMessage` already exposes all MQTT 5 properties via MqttProperties, but these are not propagated into the Pekko message model.

We would like to adopt the MQTT v5 connector long-term and would be happy to contribute an implementation to expose MQTT 5 message properties (starting with User Properties), if there are no objections or design concerns from the maintainers.

Best regards

anton-curanz-mw:
Fixes #1371

This PR adds support for MQTT User Properties in the MQTT v5 connector, enabling use-cases that rely on User Properties for metadata transport.
Since the underlying Paho client already exposes the User Properties, they simply needed to be mapped to the connector's `MqttMessage` model.
It is now possible to read User Properties from MQTT messages using `MqttSource`s as well as to publish User Properties using `MqttSink`s.

### DIFF
--- a/mqtt/src/test/travis/acl
+++ b/mqtt/src/test/travis/acl
@@ -23,6 +23,7 @@ topic v5/coffee/level
 topic v5/source-spec/will
 topic v5/source-spec/manualacks
 topic v5/source-spec/pendingacks
+topic v5/source-spec/user-props
 topic v5/sink-spec/topic1
 topic v5/sink-spec/topic2
 topic v5/sink-spec/topic3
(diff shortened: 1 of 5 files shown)


anton-curanz-mw:
Perfect, thanks for merging!
Yes it would be great to have this in 1.x as well, if that's not a problem.

Issue #1620, opened 2026-05-11
google-cloud-pub-sub-grpc: subscriber correctness follow-up to #1494


hanishi:
# google-cloud-pub-sub-grpc: subscriber correctness follow-up to #1494

## What's going on

PR #1494 added google-application-default credentials and StreamingPull auto-reconnect to the connector. After running the resulting subscriber under realistic load we hit three structural correctness gaps:

1. **(introduced in #1494)** `autoExtendAckDeadlines` doesn't actually cover messages buffered behind a slow downstream stage, so Pub/Sub redelivers them.
2. **(pre-existing since 2019, newly observable)** `GooglePubSub.subscribe(...)` propagates `maxOutstandingMessages`, `maxOutstandingBytes`, and `clientId` from the initial `StreamingPullRequest` into every keepalive request, which the server rejects with `INVALID_ARGUMENT`. The leak has been in `subscribe` since commit `3f1a4f7f28` (Jan 2019, akka.stream.alpakka era), but no caller had reason to set those fields until #1494's `RestartSettings` overload made long-lived streaming-pull subscribers viable. It's essentially a latent bug surfaced by the new use cases #1494 enables.
3. **(introduced in #1494)** `flowControlGate` is structurally identical to bug 1: it counts permits only when it pushes downstream, so its limit can't bound what the server delivers.

All three are reproducible with stub `SubscriberClient`s, no live Pub/Sub needed. See the tests at the bottom.

---

### What we expected

When you wrap a Pub/Sub source with `autoExtendAckDeadlines`, you'd expect every message that has reached your client to get its deadline extended in the background. That way Pub/Sub won't redeliver while you're still working on it.

### What actually happens

Tracking only starts when a message reaches the inside of the operator. The current code uses a plain `.map`:

```scala
Flow[ReceivedMessage]
  .via(killSwitch.flow)
  .map { msg =>
    tracked.put(msg.ackId, System.nanoTime())
    msg
  }
```

A `.map` only runs its function when downstream asks for the next element. So in a normal pipeline like this:

```scala
GooglePubSub.subscribe(request, 1.second)
  .via(GooglePubSub.autoExtendAckDeadlines(sub, 8.seconds, 30))
  .mapAsync(parallelism = 10)(processMessage)   // ~5 to 7 seconds per message
```

[...]

PR #1625, opened 2026-05-13
Google Cloud Pub/Sub gRPC: subscribe() must not echo initial-only StreamingPullRequest fields on keepalive ticks

### LINKED ISSUE #1624: google-cloud-pub-sub-grpc: subscribe() leaks initial-only fields into keepalive StreamingPullRequests, server returns INVALID_ARGUMENT
## What's wrong
`GooglePubSub.subscribe(...)` builds the keepalive request by clearing only `subscription` and `streamAckDeadlineSeconds`, leaving `clientId`, `maxOutstandingMessages`, and `maxOutstandingBytes` from the initial `StreamingPullRequest`. 
Pub/Sub doesn't allow those three on subsequent requests:

```
INVALID_ARGUMENT: You must only set max_outstanding_messages and
max_outstanding_bytes on the initial request.
```
Any caller setting `maxOutstandingMessages` to bound delivery loses the whole stream about a second after start.
## Where

`scaladsl/GooglePubSub.scala:69-71`:

```scala
val subsequentRequest = request
  .withSubscription("")
  .withStreamAckDeadlineSeconds(0)
```

Same line in `javadsl/GooglePubSub.scala`.

## Fix

One line on each DSL:

```scala
// scaladsl
val subsequentRequest = StreamingPullRequest.defaultInstance

// javadsl
val subsequentRequest = StreamingPullRequest.getDefaultInstance
```

Clears every initial-only field at once and stays safe if the proto grows more of them later.

## History

Bug has been here since 3f1a4f7f28 (Jan 2019, alpakka era), but nobody hit it until long-lived streaming-pull subscribers became viable. 
See #1620 for that broader story. Same code path is still in alpakka, akka/alpakka#3468.

This issue is a backport-friendly carve-out of the bug-2 part of #1620 so it can ship on 1.x release branches without waiting on the rest of the subscriber-correctness work.

hanishi:
`subscribe(...)` was clearing only `subscription` and `streamAckDeadlineSeconds` on the keepalive request, leaving `clientId`, `maxOutstandingMessages`, and `maxOutstandingBytes` from the initial one. Pub/Sub rejects those three on subsequent requests with `INVALID_ARGUMENT`, so any caller setting `maxOutstandingMessages` loses the whole stream about a second after start.

Fix: one line on each DSL, use `StreamingPullRequest.defaultInstance` (Scala) / `getDefaultInstance` (Java) for the keepalive. Clears every initial-only field at once and stays safe if the proto grows more later.

Bug present since `3f1a4f7f28` (Jan 2019, alpakka era). Same code path still in alpakka, akka/alpakka#3468. Broader follow-up: #1620. Regression test added in `AutoExtendAckDeadlinesSpec`.

Closes #1624.

### DIFF
--- a/google-cloud-pub-sub-grpc/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/AutoExtendAckDeadlinesSpec.scala
+++ b/google-cloud-pub-sub-grpc/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/AutoExtendAckDeadlinesSpec.scala
@@ -18,18 +18,18 @@
 package org.apache.pekko.stream.connectors.googlecloud.pubsub.grpc
 
 import org.apache.pekko
-import pekko.Done
+import pekko.{ Done, NotUsed }
 import pekko.actor.ActorSystem
-import pekko.stream.scaladsl.{ Sink, Source }
+import pekko.stream.scaladsl.{ Keep, Sink, Source }
 import pekko.stream.connectors.googlecloud.pubsub.grpc.scaladsl.{ GooglePubSub, GrpcSubscriber, PubSubAttributes }
 import com.google.protobuf.ByteString
 import com.google.pubsub.v1.pubsub._
+import org.apache.pekko.stream._
 
 import java.util.concurrent.ConcurrentLinkedQueue
 import scala.concurrent.{ Future, Promise }
 import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
-
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.concurrent.{ Eventually, ScalaFutures }
 import org.scalatest.matchers.should.Matchers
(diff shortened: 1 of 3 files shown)


google-cloud-pub-sub-grpc/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/AutoExtendAckDeadlinesSpec.scala
```diff
@@ -156,10 +158,82 @@ class AutoExtendAckDeadlinesSpec
...
+        .withMaxOutstandingBytes(10485760L)
+
+      val cancellableFut = GooglePubSub
+        .subscribe(initial, 100.millis)
+        .withAttributes(PubSubAttributes.subscriber(testSubscriber))
+        .toMat(Sink.ignore)(pekko.stream.scaladsl.Keep.left)
+        .run()
+
+      // Allow several ticks to fire.
+      Thread.sleep(500)
```
pjfanning: is there a better way to so this? Thread.sleep can be unreliable, especially on machines that have other work going on.

google-cloud-pub-sub-grpc/src/test/scala/org/apache/pekko/stream/connectors/googlecloud/pubsub/grpc/AutoExtendAckDeadlinesSpec.scala
```diff
@@ -156,10 +158,82 @@ class AutoExtendAckDeadlinesSpec
...
     system.terminate()
 }
 
+/**
+ * Stub that captures every StreamingPullRequest sent on the client → server stream
+ * and keeps the response stream open indefinitely (so the polling tick keeps firing).
+ */
+class CapturingStreamingPullClient(captured: ConcurrentLinkedQueue[StreamingPullRequest])(
+    implicit sys: ActorSystem) extends TestSubscriberClientBase {
+  private implicit val mat: pekko.stream.Materializer = pekko.stream.Materializer.matFromSystem(sys)
```
pjfanning: can you use imports to avoid fully referencing class names in code?

pjfanning (approved):
Lgtm

pjfanning:
Thanks - merged

Issue #1581, opened 2026-04-16
Fix broken FTP tests where timeouts started recently


pjfanning:
* FTP tests are failing and one possibility is that a recent set of new images for ftpd image have caused us problems
* #1580 is one failed attempt
* could be down to an upgrade in dependency libs too but the tests started failing around a time when no obvious upgrades related to FTP happened
* in short term, I think we need to disable the broken tests to keep the build clean

```
[info] FtpExamplesSpec:
[info] a file
USER *******
331 User username OK. Password required
PASS *******
--> [docs.scaladsl.FtpExamplesSpec: a file should be stored] Start of log messages of test that [Failed(org.scalatest.concurrent.ScalaFutures$$anon$1$$anon$2: A timeout occurred waiting for a future to complete. Waited 1 second.)]
16:16:15.641 INFO  [default-dispatcher-5] o.a.pekko.event.slf4j.Slf4jLogger     Slf4jLogger started
16:16:15.643 DEBUG [default-dispatcher-5] org.apache.pekko.event.EventStream    logger log1-Slf4jLogger started
16:16:15.643 DEBUG [default-dispatcher-5] org.apache.pekko.event.EventStream    Default Loggers started
16:16:15.686 DEBUG [default-dispatcher-6] zation(pekko://pekko-connectors-ftp)  Replacing JavaSerializer with DisabledJavaSerializer, due to `pekko.actor.allow-java-serialization = off`.
16:16:15.693 INFO  [ning-FtpExamplesSpec] o.a.p.s.c.t.scaladsl.LogCapturing     Logging started for test [docs.scaladsl.FtpExamplesSpec: a file should be stored]
16:16:15.709 DEBUG [default-dispatcher-5] o.apache.pekko.actor.ActorSystemImpl  Dispatcher id [pekko.stream.materializer.blocking-io-dispatcher] is an alias, actual dispatcher will be [pekko.actor.default-blocking-io-dispatcher]
16:16:16.713 INFO  [ning-FtpExamplesSpec] o.a.p.s.c.t.scaladsl.LogCapturing     Logging finished for test [docs.scaladsl.FtpExamplesSpec: a file should be stored] that [Failed(org.scalatest.concurrent.ScalaFutures$$anon$1$$anon$2: A timeout occurred waiting for a future to complete. Waited 1 second.)]
[...]

PR #1629, opened 2026-05-15
Fix unit test related to ftp timeout

christian:
This PR fixes the ftp unit test issue reported in #1581 by increasing the timeout for the test to 4 seconds (after noticing ~3s in tests):

```
[info] - should be stored (2 seconds, 883 milliseconds)
[info] - should be gzipped (2 seconds, 698 milliseconds)
```


pjfanning:
both tests failed as before - I doubt this issue is simply an issue with the length of the timeout but if you want to try increased timeouts or any other changes, go ahead

christian:
It takes a bit longer, ~4s+ but the tests are all passing now

christian:
@pjfanning could you review this when you have a chance?

He-Pin (approved):

PR #1649, opened 2026-05-29
support scala3 mongodb connector

### LINKED ISSUE #150: support mongodb for scala3
* blocked by https://jira.mongodb.org/browse/JAVA-4181
* one potential workaround is to change the compile dependencies to use mongodb-driver-reactivestreams instead of mongo-scala-driver
  * the runtime code does not use the Scala specific mongodb classes
  * the Scala specific mongodb are used in tests though - the scala tests could be removed or only run with Scala2
  * Scala 3 users who want to use the mongodb connector would need to use Java classes to bind because mongo-scala-driver has the code to support case classes, etc.

pjfanning:
https://mvnrepository.com/artifact/org.mongodb.scala/mongo-scala-driver/versions

seems like mongo-scala-driver supports Scala 3 since v5.7.0 driver

fixes #150

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -525,7 +525,7 @@ def pekkoConnectorProject(projectId: String,
       AutomaticModuleName.settings(s"pekko.stream.connectors.$moduleName"),
       mimaPreviousArtifacts := {
         if (noMimaChecks.contains(moduleName) ||
-          (moduleName == "slick" && scalaVersion.value.startsWith("3"))) {
+          (moduleName == "mongodb" && scalaVersion.value.startsWith("3"))) {
           Set.empty
         } else {
           Set(organization.value %% name.value % mimaCompareVersion)
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -392,7 +392,6 @@ object Dependencies {
       "org.apache.logging.log4j" % "log4j-to-slf4j" % Log4jVersion % Test))
 
   val MongoDb = Seq(
-    crossScalaVersions -= Scala3,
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -525,7 +525,7 @@ def pekkoConnectorProject(projectId: String,
       AutomaticModuleName.settings(s"pekko.stream.connectors.$moduleName"),
       mimaPreviousArtifacts := {
         if (noMimaChecks.contains(moduleName) ||
-          (moduleName == "slick" && scalaVersion.value.startsWith("3"))) {
+          (moduleName == "mongodb" && scalaVersion.value.startsWith("3"))) {
```
pjfanning, line 528: the old slick exclusion should have been removed ages ago but we need the mongodb exclusion for scala3 because we have no previous release jar to compare against

He-Pin (approved):

lenguyenthanh:
Can we have a `1.x` release for this or maybe just a snapshot?

pjfanning:
we publish snapshots - AI tools or Search engines will find you links - there is no current plan to support this on 1.x releases

lenguyenthanh:
I just checked, and it seems only version 2 is published with scala 3: https://repository.apache.org/content/groups/snapshots/org/apache/pekko/pekko-connectors-mongodb_3/

pjfanning:
this is not merged to the 1.x branch and there is no current plan to do this
- In Scala 3, you can use Scala 2 jars
- or you can take the code that you need from this repo and inline it in your own project and build that with Scala 3

PR #1721, opened 2026-06-26
refactor: use pattern matching instanceof (Java 16+)

He-Pin:
## Summary
- Migrate old-style `instanceof` checks with explicit casts to Java 16+ pattern matching instanceof
- 1 file updated: MqttFlowTest (1 instanceof + cast pattern migrated)
- Most files already used pattern matching; remaining instanceof checks were pure boolean tests (no cast)

## Test plan
- [ ] Existing tests pass (behavior-preserving refactoring)
- [ ] `sbt javafmtAll` passes with JDK 17


pjfanning (approved):
lgtm

PR #1733, opened 2026-06-27
one off reorder of java imports

pjfanning:
* I set `javafmtSortImports := true` in my local env
* `sbt javafmtAll` seemed to skip some of the java files where we annotate the imports for doc purposes
* the order of the imports is not configurable in sbt-java-formatter
* the existing files have pretty erratic import order - I think this PR is a bit better


He-Pin:
Will we do this to all pekko repo?

He-Pin (approved):

PR #1807, opened 2026-08-03
try to prevent traversal issues in file and ftp connectors

pjfanning:
file/src/main/scala/.../model.scala
- PathTraversalValidation utility — rejects .. segments and absolute paths
- TarArchiveMetadata — validates prefix and name on construction
- ZipArchiveMetadata — validates name on construction

ftp/src/main/scala/.../CommonFtpOperations.scala
- validatePath() — reusable validation for standalone paths
- concatPath() — validates name + checks result doesn't escape base
- retrieveFileInputStream() — validates name before FTP read
- storeFileOutputStream() — validates name before FTP write
- move() — validates both fromPath and destinationPath
- remove() — validates path before FTP delete

ftp/src/main/scala/.../SftpOperations.scala
- retrieveFileInputStream() — validates name before SFTP read
- storeFileOutputStream() — validates name before SFTP write
- move() — validates both paths before SFTP rename
- remove() — validates path before SFTP delete

### DIFF
--- a/ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
+++ b/ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
@@ -149,15 +149,19 @@ private[ftp] object CommonFtpOperations {
       s"$normPath/$normName"
     }
 
-    // Validate the normalized result doesn't escape the base path
-    val normalized = java.nio.file.Paths.get(result).normalize().toString
-    val normalizedBase = java.nio.file.Paths.get(normPath).normalize().toString
-    // On Windows, Paths.get normalizes to backslash; compare with forward-slash versions
-    val normalizedFwd = normalized.replace('\\', '/')
-    val normalizedBaseFwd = normalizedBase.replace('\\', '/')
+    // Pure string segment walk — no platform-dependent normalization.
+    // `..` and absolute names are already rejected upstream; this collapses
+    // `.` segments and empty segments (double slashes) so the prefix check
+    // cannot be fooled by cosmetic differences.
+    def collapseSegments(p: String): String = {
+      val parts = p.split('/').filter(s => s.nonEmpty && s != ".")
+      if (parts.isEmpty) "" else parts.mkString("/")
+    }
+    val collapsed = collapseSegments(result)
+    val collapsedBase = collapseSegments(normPath)
     require(
-      normalizedFwd.startsWith(normalizedBaseFwd),
-      s"concatPath result '$result' escapes base path '$normPath' after normalization")
+      collapsed == collapsedBase || collapsed.startsWith(collapsedBase + "/"),
+      s"concatPath result '$result' escapes base path '$normPath'")
 
     result
   }
--- a/ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
+++ b/ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
@@ -137,10 +137,11 @@ private[ftp] object CommonFtpOperations {
   }
 
   def concatPath(path: String, name: String): String = {
+    validatePath(name, "name")
     val normName = normalizeSeparators(name)
-    validatePath(normName, "name")
     require(!normName.startsWith("/"), s"name must not be an absolute path: '$normName'")
 
+    require(path != null, "path must not be null")
     val normPath = normalizeSeparators(path)
     val result = if (normPath.endsWith("/")) {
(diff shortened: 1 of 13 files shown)


ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
```diff
@@ -108,10 +113,41 @@ private[ftp] trait CommonFtpOperations {
...
+   * Validate that a path does not contain traversal sequences (`..`).
+   * Rejects null values and paths containing `..` as a path segment.
+   *
+   * @param path      the path to validate
+   * @param fieldName the name of the field for error messages
+   * @throws IllegalArgumentException if the path contains traversal sequences
+   */
+  def validatePath(path: String, fieldName: String): Unit = {
+    require(path != null, s"$fieldName must not be null")
+    val segments = path.split('/')
```
He-Pin: Does this work well on Windows?

pjfanning: I added a change to normalize all paths to use / but users paths with \ will be handled

pjfanning: @He-Pin could you re-review this change when you get a chance?

pjfanning: @He-Pin would you time to have a 2nd look? This is the king of issue that security researchers worry about (zip slip).

He-Pin: Will do

He-Pin (commented):
Overall the approach is solid — validating at construction time and rejecting `..` segments is the right way to handle Zip Slip / Tar Slip. Two thoughts on `concatPath` below.

He-Pin (approved):

ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
```diff
@@ -108,10 +113,52 @@ private[ftp] trait CommonFtpOperations {
...
     }
+
+    // Validate the normalized result doesn't escape the base path
+    val normalized = java.nio.file.Paths.get(result).normalize().toString
+    val normalizedBase = java.nio.file.Paths.get(normPath).normalize().toString
+    // On Windows, Paths.get normalizes to backslash; compare with forward-slash versions
+    val normalizedFwd = normalized.replace('\\', '/')
+    val normalizedBaseFwd = normalizedBase.replace('\\', '/')
+    require(
+      normalizedFwd.startsWith(normalizedBaseFwd),
```
He-Pin: `startsWith` has a sibling-prefix bypass: `"/base-evil/file".startsWith("/base")` is true. Currently unreachable since `..` is rejected upstream, but the check should be `normalizedFwd == normalizedBaseFwd || normalizedFwd.startsWith(normalizedBaseFwd + "/")` to actually hold as defense-in-depth.

ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
```diff
@@ -108,10 +113,52 @@ private[ftp] trait CommonFtpOperations {
...
+    val normPath = normalizeSeparators(path)
+    val result = if (normPath.endsWith("/")) {
+      normPath ++ normName
     } else {
-      s"$path/$name"
+      s"$normPath/$normName"
     }
+
+    // Validate the normalized result doesn't escape the base path
+    val normalized = java.nio.file.Paths.get(result).normalize().toString
```
He-Pin: `Paths.get().normalize()` is platform-dependent (Windows resolves against current drive). Since `..` and absolute names are already rejected, this block is redundant — consider a pure string segment walk instead, or drop it.

pjfanning:
It could take me a day or 2 to get back to fix the broken tests

pjfanning:
I think that I addressed the review issues. I added some CI coverage on windows for the 2 affected modules but I'm not sure if this adds much. A number of tests had to be disabled but only for the windows run.

He-Pin (approved):

ftp/src/main/scala/org/apache/pekko/stream/connectors/ftp/impl/CommonFtpOperations.scala
pjfanning: changed as suggested

PR #1809, opened 2026-08-03
mask sensitive values in toString

pjfanning:
This commit updates 11 files across multiple connector modules to mask sensitive credentials and tokens when converting objects to their string representation. The change prevents accidental exposure of secrets (API keys, tokens, passwords) in logs or debug output.

Modules affected:
- AMQP — AmqpConnectionProvider
- Google (common) — AccessToken, AccessTokenCredentials, ServiceAccountCredentials, UserAccessCredentials
- Huawei Push Kit — HmsSettings, HmsTokenApi
- IronMQ — IronMqSettings
- Jakarta Messaging — Credentials
- JMS — Credentials
- AWS S3 — model

Each file had 1–13 lines added to replace sensitive field values with masked placeholders (e.g., "***") in toString methods.

### DIFF
--- a/jms/src/main/scala/org/apache/pekko/stream/connectors/jms/Credentials.scala
+++ b/jms/src/main/scala/org/apache/pekko/stream/connectors/jms/Credentials.scala
@@ -31,7 +31,7 @@ final class Credentials private (
   override def toString =
     "Credentials(" +
     s"username=$username," +
-    s"password=${"*" * password.length}" +
+    s"password=*****" +
     ")"
 
   override def equals(other: Any): Boolean = other match {
--- a/amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/AmqpConnectionProvider.scala
+++ b/amqp/src/main/scala/org/apache/pekko/stream/connectors/amqp/AmqpConnectionProvider.scala
@@ -53,10 +53,12 @@ final class AmqpUriConnectionProvider private (val uri: String) extends AmqpConn
     factory.newConnection
   }
 
-  override def toString: String =
+  override def toString: String = {
+    val maskedUri = uri.replaceAll("://([^@]*)@", "://*****@")
     "AmqpUriConnectionProvider(" +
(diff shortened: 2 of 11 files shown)


jms/src/main/scala/org/apache/pekko/stream/connectors/jms/Credentials.scala
```diff
@@ -31,7 +31,7 @@ final class Credentials private (
   override def toString =
     "Credentials(" +
     s"username=$username," +
-    s"password=${"*" * password.length}" +
+    s"password=*****" +
```
He-Pin, line 34: Make this configurable? and I think it was better

pjfanning: giving away the length of the password is regarded as bad practice - I think hardcoding it as 5 * is ok

He-Pin (approved):

raboof (approved):
I'm in favour of this change, but it does beg the question: do we now _guarantee_ that _all_ sensitive values are masked in (for example) log output?

I think we do/should not, and should make that explicit. WDYT?

pjfanning:
> I'm in favour of this change, but it does beg the question: do we now _guarantee_ that _all_ sensitive values are masked in (for example) log output?
> 
> I think we do/should not, and should make that explicit. WDYT?

I'm not sure how we can achieve this and for avoidance of security reviewers, I think we should avoid making any explicit guarantees.

raboof:
> > I'm in favour of this change, but it does beg the question: do we now _guarantee_ that _all_ sensitive values are masked in (for example) log output?
> > I think we do/should not, and should make that explicit. WDYT?
> 
> I'm not sure how we can achieve this and for avoidance of security reviewers, I think we should avoid making any explicit guarantees.

That means we're in agreement, right? I meant making explicit that we _don't_ guarantee sensitive values are masked in (for example) log output

PR #1815, opened 2026-08-04
don't block onMessage

pjfanning:
JMS/JakartaMS JmsTxSourceStage — non-blocking acknowledgment

Problem: Await.result(envelope.commitFuture, ackTimeout) inside onMessage blocks the JMS provider's delivery thread until the user acknowledges the message. This causes thread starvation under load.

Solution: Replace blocking with deferred execution via a ConcurrentLinkedQueue:

1. onMessage no longer blocks — it drains any pending commit/rollback actions from previous messages, pushes the new envelope to the stream, and registers a commitFuture.onComplete callback that enqueues the action
2. Next onMessage call drains the queue first, executing pending actions on the provider's thread before processing the new message
3. Uses ExecutionContext.parasitic for the callback (cheap, no extra thread pool)

Files changed:
- jms/.../impl/JmsTxSourceStage.scala
- jakartams/.../impl/JmsTxSourceStage.scala

Removed: Await, TimeoutException imports, timeout/rollback handling in onMessage
Added: ConcurrentLinkedQueue, ExecutionContext.parasitic, drainPendingActions() method

### DIFF
--- a/jakartams/src/main/scala/org/apache/pekko/stream/connectors/jakartams/impl/JmsTxSourceStage.scala
+++ b/jakartams/src/main/scala/org/apache/pekko/stream/connectors/jakartams/impl/JmsTxSourceStage.scala
@@ -13,6 +13,8 @@
 
 package org.apache.pekko.stream.connectors.jakartams.impl
 
+import java.util.concurrent.ConcurrentLinkedQueue
+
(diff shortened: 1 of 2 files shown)


pjfanning:
breaks lots of tests

PR #1816, opened 2026-08-04
perf issues in google-cloud-bigquery-storage

pjfanning:
P-06 (Arrow allocator):
- SimpleRowReader now takes allocatorBytes parameter (no default)
- BigQueryStorageSettings gains arrowAllocatorBytes field (default 512MB)
- Configurable via withArrowAllocatorBytes() or config key arrowAllocatorBytes (supports size suffixes like 256 MB)
- Threaded through ArrowSource → BigQueryArrowStorage → GrpcBigQueryStorageReader

P-07 (readMerged):
- Seq.reduce((a, b) => a.merge(b)) kept — it's Seq.reduce on Source.merge which is streaming, not buffering
- Fixed SimpleRowReader leak in readRecordsMerged — now wrapped in try/finally reader.close()

### DIFF
--- a/google-cloud-bigquery-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/bigquery/storage/impl/ArrowSource.scala
+++ b/google-cloud-bigquery-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/bigquery/storage/impl/ArrowSource.scala
@@ -16,7 +16,7 @@ package org.apache.pekko.stream.connectors.googlecloud.bigquery.storage.impl
 import org.apache.pekko
 import pekko.NotUsed
 import pekko.stream.connectors.googlecloud.bigquery.storage.BigQueryRecord
-import pekko.stream.scaladsl.{ Merge, Source }
+import pekko.stream.scaladsl.Source
 import com.google.cloud.bigquery.storage.v1.arrow.{ ArrowRecordBatch, ArrowSchema }
 import com.google.cloud.bigquery.storage.v1.storage.BigQueryReadClient
 import com.google.cloud.bigquery.storage.v1.stream.ReadSession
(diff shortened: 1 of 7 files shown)


google-cloud-bigquery-storage/src/main/scala/org/apache/pekko/stream/connectors/googlecloud/bigquery/storage/impl/ArrowSource.scala
```diff
@@ -32,18 +32,24 @@ import scala.jdk.CollectionConverters._
...
-      .reduce((a, b) => a.merge(b))
+    read(client, session).reduce((a, b) => a.merge(b))
 
-  def readRecords(client: BigQueryReadClient, session: ReadSession): Seq[Source[BigQueryRecord, NotUsed]] =
+  def readRecords(client: BigQueryReadClient, session: ReadSession,
+      allocatorBytes: Long): Seq[Source[BigQueryRecord, NotUsed]] =
     read(client, session)
       .map { a =>
-        a.map(new SimpleRowReader(session.schema.arrowSchema.get).read(_))
+        a.map(new SimpleRowReader(session.schema.arrowSchema.get, allocatorBytes).read(_))
```
He-Pin: The `SimpleRowReader` created here is never closed — each batch allocates a fresh `RootAllocator` that is never released, leaking native memory over time.

`readRecordsMerged` now correctly wraps the reader in `try/finally reader.close()`, but this path does not. Consider mirroring the same pattern:

```scala
a.map { batch =>
  val reader = new SimpleRowReader(session.schema.arrowSchema.get, allocatorBytes)
  try reader.read(batch)
  finally reader.close()
}.mapConcat(c => c)
```

pjfanning: I've fixed this based on your feedback, thanks.

He-Pin:
I think our PRs should follow a format like "MMRR"—including sections such as Motivation and Modification—to help reviewers quickly grasp the intent. I will review it soon.

He-Pin (approved):

PR #1817, opened 2026-08-04
elasticsearch: escape names used in JSON requests

pjfanning:


He-Pin (approved):
Nice catch and clean fix. The old hand-rolled JSON concatenation would have produced invalid search bodies whenever a param name contained a quote or backslash, so escaping keys via spray-json is the right move while keeping values raw. I checked the rest of the elasticsearch connector for similar manual string concatenation and did not find any — the other impl classes already go through spray-json. The new spec covers the interesting cases (quotes, backslashes, newlines, empty map), and CI is green. Approving.

raboof (approved):
This is a good improvement, though this API remains quite a footgun. The documentation of where these maps are passed in to the public API do seem to document the fact that the values are raw JSON strings correctly, so I think we're 'fine', but it would be a nice follow-up improvement to introduce APIs that take `JsObject` values and similar (and if that works out, possibly even deprecate the old string-based ones and provide them only through an API method with 'Raw' explicitly in the method name).

PR #1819, opened 2026-08-04
s3: discourage use of static credentials

pjfanning:


pjfanning:
There is a CI issue with the kudu tests not starting but I don't think this blocks reviewing and merging this s3 change.

He-Pin (approved):

PR #1857, opened 2026-08-22
unix-domain-socket: replace deprecated use of Source.queue

pjfanning:
UnixDomainSocketImpl has no deprecated Source.queue usage left, and SourceQueueWithComplete is gone from the file entirely.

Receive side → Source.actorRefWithBackpressure. The ack protocol maps exactly onto the existing OP_READ gating: pekko sends the ack only after push(out, e), i.e. once the stream has actually taken the element — the same guarantee QueueOfferResult.Enqueued gave. The io thread's state machine is untouched in shape:

- ReceiveQueue.offer(bytes) sets a Promise[Done], then ref.tell(bytes, ackReceiver), returning the future the loop already polls via pendingResult.isCompleted.
- A per-connection ReceiveAckActor receives the ack, completes that promise, and calls sel.wakeup() — the same wake-up the old pendingResult.onComplete did. The actor only completes a promise; the io thread remains the sole mutator of context.receive, preserving the existing threading discipline.
- complete() sends ReceiveComplete, matched by the completion matcher as CompletionStrategy.draining. I checked ActorRefBackpressureSource: the completion matcher is tested before the element branch, so the send-side's receiveQueue.complete() is safe even while an ack is outstanding.
- queue.watchCompletion() → a watchTermination(Keep.both) future carried on ReceiveQueue.completion.
- Source.futureSource(...) and the prefixAndTail(0)/Sink.head pre-materialisation trick are replaced by preMaterialize().

One failure mode the old code got for free needed explicit handling: if the read-side stream terminates (downstream cancels), no ack will ever arrive, so the io thread would sit forever with OP_READ cleared. The termination hook now fails any outstanding ack promise, which drives the loop's existing shutdown branch, stops the ack actor, and wakes the selector.

Verified: unix-domain-socket/test — 6 passed, 1 ignored, Java test passes; MiMa clean (the class is under impl, covered by the *.impl.* filter); scalafmt applied and clean.

### DIFF
--- a/unix-domain-socket/src/main/scala/org/apache/pekko/stream/connectors/unixdomainsocket/impl/UnixDomainSocketImpl.scala
+++ b/unix-domain-socket/src/main/scala/org/apache/pekko/stream/connectors/unixdomainsocket/impl/UnixDomainSocketImpl.scala
@@ -245,7 +245,7 @@ private[unixdomainsocket] object UnixDomainSocketImpl {
 
   private def acceptKey(
       localAddress: JnrUnixSocketAddress,
-      incomingConnectionQueue: SourceQueueWithComplete[IncomingConnection],
+      incomingConnectionQueue: BoundedSourceQueue[IncomingConnection],
       halfClose: Boolean,
       receiveBufferSize: Int,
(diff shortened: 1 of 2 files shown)


pjfanning:
@He-Pin does this change look ok?

He-Pin (approved):

PR #1895, opened 2026-09-02
HBase: re-enable integration tests in CI

### LINKED ISSUE #61: hbase CI build is disabled

pjfanning:
### Motivation
The hbase CI job has been disabled since 2020 (inherited from alpakka), so the HBase connector has no integration coverage, and the connector is stuck on the EOL `hbase-shaded-client` 1.4.14 from 2020 — the reason `org.apache.hbase`/`org.apache.hadoop` are frozen in `.scala-steward.conf`. The root cause in akka/alpakka#2185 was that the `harisekhon/hbase` container's region server registers itself in ZooKeeper under the hostname `hbase`, which the runner cannot resolve — all tests then time out.

### Modification
- Add `scripts/hbase-server.sh` (same pattern as `scripts/ftp-servers.sh`, since `$PRE_CMD` cannot contain shell operators): appends `127.0.0.1 hbase` to `/etc/hosts` and starts the `hbase` docker-compose service; re-enable the `hbase` entry in the CI connectors matrix.
- Publish the HBase-default ports 16020/16030 instead of the stale 16201/16301 (the 2020 image rebuild moved the region server to the default port).
- Use the `harisekhon/hbase:2.1` image (HBase 2.x server).
- Upgrade the client to `hbase-shaded-client`/`hbase-common` 2.6.6 (current release line); migrate `HBaseCapabilities` off the removed-in-3.0 `HTableDescriptor`/`HColumnDescriptor` to `TableDescriptorBuilder`/`ColumnFamilyDescriptorBuilder`, and the test snippets off deprecated `Append.add` to `Append.addColumn`.
- Remove the `org.apache.hbase`/`org.apache.hadoop` ignores from `.scala-steward.conf` so both are steward-managed again.

### Result
The HBase connector has working CI integration tests again, on a current HBase client, with hbase/hadoop dependency updates no longer frozen.

### Tests
- The re-enabled `connectors (hbase)` CI job runs the full suite (passed against both the 1.4 client + 2.1 server and now validates the 2.6.6 client).
- Locally: `sbt hbase/Test/compile` clean (no deprecation warnings), `sbt hbase/mimaReportBinaryIssues` passes, `scalafmt --list --mode diff-ref=upstream/main` clean.

### References
Fixes #61, Refs https://github.com/akka/alpakka/issues/2185


pjfanning:
I tested hbase with same lib version and docker image version at the start which indicates that the 1.x releases of pekko-connectors-hbase are likely to work ok. As a follow up, I may create a 1.x PR to re-enable hbase testing there.

nvollmar (approved):

PR #1928, opened 2026-09-06
Google: close the credentials file when parsing it fails

pjfanning:
`ServiceAccountCredentials` and `UserAccessCredentials` both read their JSON file like this:

```scala
val src = Source.fromFile(c.getString("path"))
val credentials = JsonParser(src.mkString).convertTo[ServiceAccountCredentialsFile]
src.close()
```

`close()` is not in a `finally`, so a parse failure skips it and the file descriptor leaks.

That is not a rare path — it is the expected one. The `application-default` provider tries `parseServiceAccount` first inside a `NonFatal` catch, and on any machine set up with `gcloud auth application-default login` the well-known file is an *authorized user* file. So `convertTo[ServiceAccountCredentialsFile]` throws, the descriptor leaks, and the cascade falls through to the user access parser, which reads the same file successfully.

One leaked descriptor per `Credentials` construction on the standard developer path. `GoogleExt` caches settings per config path per `ActorSystem`, so it is bounded in normal use, but `GoogleSettings(config)` is public and un-cached, so building credentials per tenant or per request scope leaks one each time.

Both sites now close in a `finally`.

### Testing

Four cases in `CredentialsSpec`: each parser reading its own file format, and each parser being handed the *other* format, which is the case that used to leak. The failure cases assert the exception still propagates, so the `finally` is not swallowing it.

There is no portable way to assert descriptor closure from a JVM test, so these are behavioural guards around the change rather than a direct assertion that the handle was released. `google-common/test` passes (28) and `checkCodeStyle` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/google-common/src/test/scala/org/apache/pekko/stream/connectors/google/auth/CredentialsSpec.scala
+++ b/google-common/src/test/scala/org/apache/pekko/stream/connectors/google/auth/CredentialsSpec.scala
@@ -25,6 +25,9 @@ import org.scalatest.BeforeAndAfterAll
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpecLike
 
+import java.nio.charset.StandardCharsets.UTF_8
+import java.nio.file.{ Files, Path }
+
 class CredentialsSpec
     extends TestKit(ActorSystem("CredentialsSpec"))
     with AnyWordSpecLike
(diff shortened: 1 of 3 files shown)


google-common/src/test/scala/org/apache/pekko/stream/connectors/google/auth/CredentialsSpec.scala
```diff
@@ -36,6 +39,78 @@ class CredentialsSpec
...
+      val config = ConfigFactory.parseString(s"""
+           |project-id = ""
+           |client-email = ""
+           |private-key = ""
+           |path = "$path"
+        """.stripMargin)
+
+      // this is the routine case: `gcloud auth application-default login` writes a user access file,
+      // and the application-default provider tries the service account parser against it first
+      a[Exception] should be thrownBy ServiceAccountCredentials(config, Set("a-scope"))
```
Philippus:
```suggestion
      an[Exception] should be thrownBy ServiceAccountCredentials(config, Set("a-scope"))
```

google-common/src/test/scala/org/apache/pekko/stream/connectors/google/auth/CredentialsSpec.scala
```diff
@@ -36,6 +39,78 @@ class CredentialsSpec
...
+    withJsonFile(serviceAccountJson) { path =>
+      val config = ConfigFactory.parseString(s"""
+           |project-id = ""
+           |client-id = ""
+           |client-secret = ""
+           |refresh-token = ""
+           |path = "$path"
+        """.stripMargin)
+
+      a[Exception] should be thrownBy UserAccessCredentials(config)
```
Philippus:
```suggestion
      an[Exception] should be thrownBy UserAccessCredentials(config)
```

Philippus (changes requested):
2 nitpicks.

Philippus (approved):

PR #1935, opened 2026-09-09
Elasticsearch: delete test indices by name instead of _all

pjfanning:
### Motivation
Every Elasticsearch/Opensearch test suite wipes its server between runs with `DELETE /_all`, and none of them check the response.

Elasticsearch 8 [changed `action.destructive_requires_name` to default to `true`](https://www.elastic.co/guide/en/elasticsearch/reference/8.19/indices-delete-index.html), which refuses wildcard and `_all` destructive operations. So from Elasticsearch 8 on, the cleanup silently does nothing — the request is rejected, the response is discarded, and the suite carries on.

The consequence is not a cleanup failure but a confusing failure somewhere else entirely. Documents survive into the next suite, which reads them back and cannot deserialize them: the Java `Book` has only a `title`, while the Scala `Book` also carries a `price`, so the Java suite fails with `UnrecognizedPropertyException: Unrecognized field "price"` on an index it believed it had just created. This was hit for real on #1934, which adds Elasticsearch 8 and 9 test servers.

The suites are not racing — `Test / parallelExecution := false` applies to every connector project, so they run one at a time and `_all` deletion between them is the intended isolation. It just stopped working.

### Modification
- Replaced the `_all` deletions with a helper that lists indices via `_cat/indices?h=index` and deletes them by explicit name, which every supported server accepts. Index names starting with `.` are left alone.
- Both the list and the delete response are now checked, so a cleanup that stops working fails the suite instead of passing quietly.
- Added as `deleteAllIndices` in `ElasticsearchSpecUtils` for the four Scala suites, and rewrote `ElasticsearchTestBase.cleanIndex` for the Java ones.
- Renamed the Scala `sink3-0` index to `sink3-0-scala` — the last index name shared between a Java and a Scala suite running against the same server. Harmless while cleanup works, but it is the exact trap that produced the failure above.
- Dropped the imports left unused by the change.

Test-only; no main source touched.

### Result
Cleanup works against every server the tests run against, including Elasticsearch 8 and 9, and a future regression in it surfaces as a cleanup failure rather than as a deserialization error in an unrelated suite.

### Tests
- `sbt "elasticsearch/Test/compile"` — success, no warnings
- `sbt "++3.3.8 elasticsearch/Test/compile"` — success (Scala 3)
- `sbt "elasticsearch/scalafmtAll" "elasticsearch/javafmtAll"` — success
- Integration run not done locally (no Docker on this machine); relying on the `elasticsearch` CI job.

### References
Refs #1934 — this is the underlying cause of the failure seen there

### DIFF
--- a/elasticsearch/src/test/java/docs/javadsl/ElasticsearchTestBase.java
+++ b/elasticsearch/src/test/java/docs/javadsl/ElasticsearchTestBase.java
@@ -16,10 +16,12 @@ package docs.javadsl;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.pekko.actor.ActorSystem;
 import org.apache.pekko.http.javadsl.Http;
 import org.apache.pekko.http.javadsl.model.ContentTypes;
 import org.apache.pekko.http.javadsl.model.HttpRequest;
+import org.apache.pekko.http.javadsl.model.HttpResponse;
 import org.apache.pekko.stream.connectors.elasticsearch.ApiVersion;
 import org.apache.pekko.stream.connectors.elasticsearch.ApiVersionBase;
 import org.apache.pekko.stream.connectors.elasticsearch.ElasticsearchConnectionSettings;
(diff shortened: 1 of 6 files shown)


pjfanning:
added to #1934 because its CI builds kept failing

